PR #32, opened 2026-05-24
Remove the qwen context size of 4096 since it is very configurable these days

gherlein:
**Change**: Removed the entry `"qwen": 4_096,` from the `contextWindowSizes` map (line 224)    
   - **Effect**: This line was preventing qwen models from being recognized by the provider system,     
which was causing context window information to not display correctly. By removing this hard-coded   
entry, qwen models can now be properly resolved to their actual provider (Ollama, OpenAI compatible,    
etc.) and their correct context window sizes will be determined through the provider resolution logic    
rather than being incorrectly hardcoded to 4,096 tokens.

Also added a few documentation files helpful for new users.


gherlein:
Am going to remove the docs to make this a cleaner PR

PR #56, opened 2026-07-20
test: isolate HOME so the suite cannot overwrite a real ~/.pi-go

bclermont:
Running `go test ./...` overwrites the developer's own `~/.pi-go` configuration.

Test code reaches `saveThemeToConfig` and `SaveDefaultRole`, which resolve `~/.pi-go/config.json` via `os.UserHomeDir` and marshal the **entire** `Config` back to disk. Fixture state therefore gets persisted over real configuration. On my machine the `default` role was replaced with `claude-sonnet-4-6` — a model I have no API key for — and `pi` failed to start until I repaired the file by hand:

```
Error: no API key found for provider "anthropic" (set ANTHROPIC_API_KEY)
```

The blast radius is wider than `config.json`. Running the suite under a sacrificial `HOME` shows what a full run writes:

```
$HOME/.pi-go/config.json
$HOME/.pi-go/history.jsonl
$HOME/.pi-go/last-session.json
$HOME/.pi-go/usage.json
$HOME/.pi-go/sessions/...
$HOME/.pi-go/log/...
$HOME/.pi-go/memory/claude-mem.db
$HOME/.pi-go/models/sentence-transformers_all-MiniLM-L6-v2/   (~90MB download)
```

Some tests already isolate `HOME` individually — which is why this went unnoticed. The ones that forget are the leak, and nothing enforces it.

## Change

A package-scoped `TestMain` in each package whose code can reach a `$HOME` write: `cli`, `tui`, `agent`, `sop`, `guardrail`. Isolation then holds by construction rather than by every future test remembering. Tests that need their own `HOME` still override it — `t.Setenv` restores the throwaway directory instead of the developer's.

`internal/tui` already had a `TestMain` for browser/restart stubbing, so it is extended rather than duplicated.

`lastSessionFile` needs separate handling: it is a package-level var resolved from `$HOME` at **init**, which runs before `TestMain`, so setting the environment cannot reach it. `TestMain` repoints the var directly.

## Verification

Full suite run under a sacrificial `HOME`, before and after, diffing what appeared:

| | before | after |
|---|---|---|
| paths created under `~/.pi-go/` | 8 (incl. a ~90MB model download) | **0** |

`go test ./...` passes, `gofmt -l` and `go vet ./...` clean. No production code changed — the diff is test files only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/agent/main_test.go
+++ b/internal/agent/main_test.go
@@ -1,0 +1,35 @@
+package agent
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+// TestMain redirects HOME to a throwaway directory for every test in this
+// package.
+//
+// Code under test resolves paths through os.UserHomeDir and writes there —
+// ~/.pi-go/config.json, session history, logs. Tests that exercise those paths
+// without isolating HOME first will overwrite the developer's real
+// configuration: running "go test ./..." was enough to rewrite the default
+// model role and theme of the machine running it. Isolating at package scope
+// makes that impossible by construction, rather than relying on every future
+// test remembering to call t.Setenv("HOME", ...).
+//
+// Tests that need their own HOME can still override it; t.Setenv restores this
+// value rather than the developer's when they finish.
+func TestMain(m *testing.M) {
+	dir, err := os.MkdirTemp("", "pi-go-test-home-")
+	if err != nil {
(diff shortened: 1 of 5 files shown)


bclermont:
maybe you should use https://github.com/spf13/afero instead and you will save a lot of time, especially with jailing tools call to root folder

PR #92, opened 2026-08-08
fix(tools): stop leaking process trees, stream shell output, let the agent cancel

dimetron:
## What broke

A `bash` tool call could hang forever.

`exec.CommandContext` signals only the **direct child** on cancellation, and `cmd.Wait()` blocks until the stdout/stderr pipes reach EOF. A compound command — a pipeline, or anything with `;` — forks grandchildren that inherit those pipes. Killing the shell leaves them running, reparented to init, still holding the write end. `Wait` then never returns, and `WaitDelay` was zero so nothing forced it.

Two sessions sat wedged for over an hour this way:

| Session | Command | Stuck |
|---|---|---|
| `260808-0714-51211-aff55` | `find / -path '*genai*' -name '*.go' \| grep -i live` | 1h20m |
| `260808-0714-de8bc-c1017` | `find / -type d -path '*adk/v2*'` | 1h20m |

`lsof` confirmed the mechanism: `find` (PPID 1) still owned fd 1 of pi's pipe. Neither session ever received a tool result, so neither could advance. The 2-minute timeout was unreachable.

**Not a regression.** `git log --follow internal/tools/bash.go` shows the same shape in all 11 commits back to the initial version — `Setpgid`, `cmd.Cancel` and `WaitDelay` have never been set. Only the trigger is new: the hang needs a grandchild that outlives the shell *and* a command slow enough to hit the timeout.

## The fix

**`internal/procs`** — every child leads its own process group; `Cancel` sends `SIGTERM` to the group and `SIGKILL` after a grace period; `WaitDelay` is a backstop for anything group termination cannot reach (a stopped process, a descendant that escaped into a new session, a platform without process groups). `procs.CommandContext` is the constructor, because `Isolate` on a non-context command fails only at `Start` time.

The same defect existed in two more places, fixed here as well:
- `internal/extension/hooks.go` — both hook exec paths
- `internal/tui/run.go` — `/run` gate commands

**Timeouts no longer kill.** A command that outruns its deadline, or that goes fully silent past the idle threshold, is handed to a supervisor and reported with a handle while it keeps running:

```
bash("find / -name '*.go'")
  → { running: true, handle: "bg_3", stdout: "<partial>",
      note: "Command no output for 90s and was moved to the background; it is
             still running. Read more with bash_output(handle=\"bg_3\") or stop
             it with bash_kill(handle=\"bg_3\"). It has produced no output at
             all — if that is unexpected, the command is probably too broad …" }
```

Two new tools: `bash_output` (incremental read, optional blocking wait) and `bash_kill` (stops the whole tree). Killing on timeout threw away work and explained nothing; handing off loses neither.

**Idle threshold: 90s**, calibrated rather than picked round — a cold `go build ./...` on this repo prints nothing for **28s**, so a 30s threshold would background ordinary builds. Tunable per call via `idle_timeout`.

**Live output.** Command output streams to the main window as it is produced, with a state line that makes a stall visible rather than indistinguishable from a hang:

```
◉ bash(find / -name '*.go')
  │ /usr/local/go/src/…
  │ ⏳ no output for 90s
```

Events share the subagent channel under a `bash:` prefix and bind to their card **by command text**, not by position — the model routinely issues several bash calls per turn, and position matching would interleave them.

**The handoff cannot become a new leak:** background commands are capped (8, evicting finished ones first), lifetime-limited (30 min), killed by `bash_kill`, and killed at session shutdown via `KillAll` in both CLI paths. Foreground cancellation (Esc) still kills immediately and never backgrounds.

## Tests

- `procs`: reproduces the exact wedge — a command that backgrounds a pipe-holding child, cancelled; asserts `Wait` returns **and** the grandchild is dead. Plus normal-exit, explicit-kill, and kill-after-exit paths.
- `tools`: silent command backgrounds; **chatty command does not** (the guard against an over-eager idle check); incremental reads don't repeat or lose output; handle spent after exit; foreground cancel kills without backgrounding; `KillAll`; cap eviction; sink streams stdout/stderr/heartbeat/stall.
- `tui`: binding by command text across concurrent calls, finished cards not rebound, event cap, stall state rendered, result replaces the live stream.
- Stream buffer: absolute offsets, dropped-byte reporting, wake-on-write, concurrent writers.

`go test ./...` clean, `-race` clean on all touched packages, `make lint` 0 issues.

## Note

The three orphans from those two sessions were killed manually while diagnosing.

### DIFF
--- a/internal/acp/server/runtime.go
+++ b/internal/acp/server/runtime.go
@@ -327,15 +327,25 @@ func buildSessionResources(rt RuntimeConfig, cfg config.Config, turn PromptTurn,
 		_ = sandbox.AddExtraDir(filepath.Join(home, ".pi-go"))
 	}
 
-	coreTools, err := tools.CoreTools(sandbox)
+	bashSup := tools.NewBashSupervisor()
+	coreTools, err := tools.CoreTools(sandbox, tools.WithBashSupervisor(bashSup))
 	if err != nil {
 		_ = sandbox.Close()
 		return nil, fmt.Errorf("creating core tools: %w", err)
 	}
+	bashCtlTools, err := tools.BashControlTools(bashSup)
+	if err != nil {
+		_ = sandbox.Close()
+		return nil, fmt.Errorf("creating bash control tools: %w", err)
+	}
+	coreTools = append(coreTools, bashCtlTools...)
 
 	orch := newSessionOrchestrator(rt, cfg, cwd)
 	lspMgr := lsp.NewManager(nil)
 	cleanup := func() {
+		// Backgrounded commands outlive the turn by design, so the session
+		// teardown is the only thing that will ever reap them.
+		bashSup.KillAll()
 		lspMgr.Shutdown()
 		orch.Shutdown()
 		_ = sandbox.Close()
--- a/internal/cli/cli.go
+++ b/internal/cli/cli.go
@@ -465,6 +465,7 @@ type nonInteractiveRuntime struct {
 	coreTools    []adktool.Tool
 	orch         *subagent.Orchestrator
 	agentEventCh chan tui.AgentSubEvent
+	bashSup      *tools.BashSupervisor
 }
 
 func initNonInteractiveRuntime(ctx context.Context, cfg *config.Config, cwd, sandboxRoot, worktreeDir string) (*nonInteractiveRuntime, error) {
@@ -479,11 +480,18 @@ func initNonInteractiveRuntime(ctx context.Context, cfg *config.Config, cwd, san
 		}
 	}
 
-	coreTools, err := tools.CoreTools(sandbox)
+	bashSup := tools.NewBashSupervisor()
+	coreTools, err := tools.CoreTools(sandbox, tools.WithBashSupervisor(bashSup))
 	if err != nil {
 		_ = sandbox.Close()
 		return nil, fmt.Errorf("creating core tools: %w", err)
 	}
+	bashCtlTools, err := tools.BashControlTools(bashSup)
+	if err != nil {
+		_ = sandbox.Close()
+		return nil, fmt.Errorf("creating bash control tools: %w", err)
+	}
+	coreTools = append(coreTools, bashCtlTools...)
 
 	repoRoot := detectGitRoot(ctx, cwd)
 	discovery, err := subagent.DiscoverAgents(cwd, subagent.ScopeBoth)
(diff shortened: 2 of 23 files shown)


dimetron:
## Update: coverage + a third wiring site + end-to-end verification

### Coverage

| Package | Before | After |
|---|---:|---:|
| `internal/procs` (new) | 81.8% | **90.9%** |
| `internal/tools` | 85.2% | **86.4%** |
| `internal/tui` | — | 89.0% |
| `internal/extension` | — | 87.1% |
| **Repo total** | — | **86.4%** |

Every function this branch adds is now ≥84.6%; the two still under 90% are `procs.terminate` (80%, the unreachable double-signal-failure branch) and `BashControlTools` (84.6%, its two `newTool` error returns).

The new tests target the gaps that matter rather than the ones that were easy:

- **The kill(2) sign guard.** `signalGroup` falls back to the direct child when no process group was established. Getting that wrong means `kill(-pid)` names the group *pi itself* is in and takes down the session — so the test asserts both that the child dies and that the test process survives.
- **SIGTERM escalation** against `trap '' TERM; sleep 30` — a shell that ignores TERM must still be reaped.
- **No-ops on unstarted/reaped commands.** `terminate` runs on `Wait`'s goroutine; a panic there would break the exact cancellation path this PR fixes.
- **Both control tools through the ADK `Run` path**, so argument decoding is covered rather than just the supervisor underneath.
- **The lifetime reaper** and **eviction preferring finished commands over live ones**.
- **`WithBashSupervisor` actually reaching the bash tool** — if it didn't, the CLI would attach its sink to a supervisor nothing runs through and streaming would silently never appear.

**One package is below target and it is pre-existing:** `internal/cli` is 80.4% here vs **80.5% on `main`** — untouched by this branch beyond the wiring, whose new lines land in `cleanup` (100%) and `close` (100%). Raising it is a separate job. Same for `httplog` (72.8%), `session` (77.7%), `guardrail` (81.8%), `palace` (81.6%), `codex` (82.0%), `jsonrpc` (84.9%).

### A third wiring site I had missed

`internal/acp/server/runtime.go` built its tools through the unwired `CoreTools` path. ACP sessions got the leak fix, but **no** `bash_output`/`bash_kill`, no streaming sink, and nothing killing background commands at teardown. Now wired to the same supervisor as both CLI paths — all three `CoreTools` callers are covered.

### End-to-end verification

Not just unit tests — a real session against `minimax-m3:cloud`:

```
⚙ bash      {"command":"sleep 300","idle_timeout":3000}
✓ bash done
⚙ bash_kill {"handle":"bg_1"}
✓ bash_kill done
→ Handle: bg_1 — yes, killed (after ~6.8s, exit code -1).
```

`ps` afterwards: no orphaned `sleep`. The model discovered the handle, used it, and the process group died with it.

`go test ./...` clean, `make lint` 0 issues, `-race` clean on all touched packages.

PR #112, opened 2026-08-09
feat(provider): send Ollama repetition-control sampling options

dimetron:
## What

pi-go sent exactly one Ollama option, `num_predict`. The four repetition controls Ollama exposes ran at server defaults with no way to tune them, so a model in repetition collapse had no knob at all.

| Env var | Ollama option | Ollama default |
|---|---|---|
| `PI_OLLAMA_REPEAT_PENALTY` | `repeat_penalty` | `1.1` |
| `PI_OLLAMA_REPEAT_LAST_N` | `repeat_last_n` | `64` |
| `PI_OLLAMA_PRESENCE_PENALTY` | `presence_penalty` | `0.0` |
| `PI_OLLAMA_FREQUENCY_PENALTY` | `frequency_penalty` | `0.0` |

## Design decisions

- **Opt-in, no new defaults.** An unset var is omitted from the request entirely rather than sent as a zero value, so Ollama's defaults stay in force and behaviour is byte-identical for anyone who sets nothing. Picking defaults would change generation quality for every Ollama model, and a value that helps one can degrade another.
- **`0` and `-1` are forwarded, not swallowed.** Both are meaningful to `repeat_last_n` — `0` disables the penalty ring, `-1` means `num_ctx` — so "set to 0" must not be indistinguishable from "unset".
- **Invalid values are ignored, not fatal.** A typo in an env var should not take down a session that would otherwise run; a valid sibling still applies.
- **One request path.** Local Ollama and Ollama Cloud share it, so these apply to both.

## Why

`num_predict` bounds how far a degenerate turn runs; it does not stop the turn degenerating. Ollama penalises repeats across the last 64 tokens by default, while measured degenerate turns on `deepseek-v4-flash` cycle on phrases of 89–194 bytes (~25–55 tokens) — so a full cycle can fall outside the window the penalty can see.

Found via a sweep of 496 session logs (29 days, 25 models). Degenerate turns were concentrated almost entirely in `deepseek-v4-flash` on Ollama: other models peaked at a longest tool-free thinking run of 0–45 events, DeepSeek reached 164 events / 148 KB.

## Validation

Method: resume a seed session whose last persisted event is the tool result immediately before a known spiral, then score the resulting log. Wire format confirmed live: `"options":{"num_predict":16384,"repeat_last_n":512,"repeat_penalty":1.2}`.

| Configuration | Trials | Looped | byte-exact `reps` |
|---|---|---|---|
| Baseline, ollama-local | 3 | **3/3** | 0, 67, 42 |
| Baseline, ollama-cloud | 3 | **2/3** | 1, 71, 1 |
| `repeat_last_n=512`, `repeat_penalty=1.2` | 3 | **1/3** | **0, 0, 0** |

**The penalties do what they are designed to do, and no more.** Byte-exact repetition was eliminated. But the surviving failure still spun 57 consecutive thinking events / 23 KB with no tool call — the model stopped repeating *verbatim* and kept repeating *semantically*.

Two consequences, both stated in the spec:

1. The 3/3 → 1/3 rate drop is suggestive but **not established** at n=3 with unpinned sampling. The `reps` → 0 effect is the solid part.
2. These settings can make the failure **harder to detect**. `observeOutput` keys on byte-exact periodicity, so suppressing verbatim repeats can convert a detectable loop into a silent one.

**No default is proposed on this evidence.** A default should land together with a semantic-loop detector, or it trades a visible failure for an invisible one.

## Tests

6 new in `internal/provider/ollama_sampling_test.go`:

- unset sends nothing (Ollama defaults preserved)
- values marshal with correct types
- garbage dropped while a valid sibling survives
- `0` and `-1` preserved as meaningful
- wire test: merge does not clobber `num_predict`
- regression: no sampling keys present when unset

Full unit suite green, `go vet` clean, `golangci-lint run internal/provider/...` → 0 issues.

> Note: `TestNewPromptHandler_NoAPIKey` (`internal/acp/server`) fails when `ANTHROPIC_API_KEY` is set in the environment — it expects the no-key path but makes a real API call. Confirmed **pre-existing on clean `main`** and passing with the key unset. Unrelated to this change.

## Also included

- **`.pi-go/skills/pi-loop-forensics/`** — the skill used to produce these measurements: `scan_logs.py` (corpus sweep), `score_run.py` (per-log scoring), `ab_replay.sh` (provider A/B replay with preflight). Documents the detector's mechanics, a commit-timeline caveat for judging old logs, a three-way procedure for discriminating a race vs a tool-parse failure vs a too-low guard, and the fact that `pi ping` resolves URLs and credentials differently from the real agent path (it gave three false "provider down" readings).
- **Spec correction** — `specs/features/LLM/002-ollama-sampling-parameters` had the `repeat_last_n` sentinels inverted, claiming `0` scans the whole context. Verified against `ollama@v0.32.5` (`x/mlxrunner/sample/sample.go:225`): `0` **disables** the penalty entirely, overriding the coefficients, and `-1` means `num_ctx`.

## Follow-up, not in this PR

The runaway guard (`stuckDetector`) is instantiated only in `internal/tui/agent_loop.go`. `runPrint` and the json/socket/rpc and ACP paths have **no repetition protection at all** — replay trials with 71, 67 and 42 byte-exact repeats ran to completion without aborting. There is also no session-level token cap; the only `TokenBudget` in the tree belongs to the memory subsystem.

### DIFF
--- a/.pi-go/skills/pi-loop-forensics/SKILL.md
+++ b/.pi-go/skills/pi-loop-forensics/SKILL.md
@@ -1,0 +1,168 @@
+---
+name: pi-loop-forensics
+description: Diagnose pi-go agent loops and degenerate turns — "agent loop aborted", runaway thinking with no tool calls, repeated phrases. Discriminates genuine model repetition collapse from a race, a tool-parse failure, or a too-low guard, and A/B replays a seed session across providers.
+---
+
+# Loop Forensics
+
+Use when a run dies with `agent loop aborted: ...`, or a model burns a long turn
+thinking without ever calling a tool. This skill decides **why** before anything
+is changed: the abort message names a symptom, not a cause.
+
+Sibling skill `pi-check-session-logs` covers *tool call errors*. This one covers
+*repetition and runaway turns*. They do not overlap.
+
+## Where the evidence lives
+
+| What | Path |
+|---|---|
+| Session logs (JSONL, one per run) | `~/.pi-go/log/<yyyy-mm-dd>/session-HH-MM-SS.log` |
+| Session state (resumable) | `~/.pi-go/sessions/<id>/` — `events.jsonl`, `meta.json`, `trajectory.atif.json` |
+| Detector source | `internal/tui/agent_loop.go` |
+
+Log rows are `{"time","type","content",...}` with `type` in `session_start`,
+`thinking`, `llm_text`, `tool_call`, `tool_result`, `user`, `error`. The model
+name is on the `session_start` row — always record it, findings are per-model.
+
+## How the guard actually works
+
+`stuckDetector` (`internal/tui/agent_loop.go:119`) has three independent arms:
+
+- `observe` — identical consecutive tool calls, trips at `maxRepeatToolCalls`=10.
+  Pagination args are stripped first (`volatileToolArgs`), so paging one file
+  collapses to one fingerprint.
+- `observeError` — same tool failing `maxToolErrorStreak`=10 times running.
+- `observeOutput` (`:246`) — the model's own text/thinking. Needs a period ≥
+  `minOutputPeriod`=16 bytes, ≥ `minPeriodVariety`=8 distinct bytes, and
+  `maxOutputRepeats`=12 **byte-exact** back-to-back copies inside an 8 KB tail.
+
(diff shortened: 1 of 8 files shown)


dimetron:
https://github.com/ollama/ollama/issues/17617

PR #165, opened 2026-08-14
feat(piagent): public package for embedding a PiAgent

dimetron:
## What this adds

A top-level `piagent/` package — the module's only importable surface, since
every other package is under `internal/`. It assembles the same agent the `pi`
CLI runs headlessly (`--mode print`), minus the TUI.

**Provider construction is deliberately not here.** `piagent` never resolves a
provider, looks up a credential or picks an endpoint. The model arrives through
`WithModel` as an ADK `model.LLM`, and `New` returns `ErrNoModel` without one.
It composes with `pimodels` (#164) in the embedder's code:

```go
m, err := pimodels.FromConfig(ctx, "")            // the model a pi session picks
ag, err := piagent.New(ctx, piagent.WithModel(m)) // the agent
defer ag.Close()
```

Two lines instead of one is the honest price of the boundary.

`piagent/isolation_test.go` makes that a CI gate, not a doc comment: it fails
if `piagent` ever reaches `internal/provider`, `internal/guardrail` or
`pimodels`, and fails if it stops depending on ADK's `model` package — the
other half of the contract, since the two packages have to meet at a type
neither owns. Verified by temporarily importing `internal/provider` and
watching it fail.

### Exported surface

```go
func New(ctx context.Context, opts ...Option) (*Agent, error)
var ErrNoModel error

func (a *Agent) Close() error
func (a *Agent) NewSession(ctx context.Context) (string, error)
func (a *Agent) Run(ctx, sessionID, message string) iter.Seq2[*session.Event, error]
func (a *Agent) RunStreaming(ctx, sessionID, message string) iter.Seq2[*session.Event, error]
func (a *Agent) Ask(ctx, sessionID, message string) (string, error)
func (a *Agent) SetSessionTitle(sessionID, title string) error
func (a *Agent) Tools() []tool.Tool
func (a *Agent) Model() string
func (a *Agent) WorkingDir() string

type LSPMode string  // LSPOff, LSPMin, LSPFull
type AgentEventFunc func(id, kind, content string)
type Option func(*options)
```

| Option | Purpose |
|---|---|
| `WithModel(m model.LLM)` | **Required.** The model the agent runs on |
| `WithWorkingDir(dir)` | Sandbox root and discovery anchor |
| `WithExtraSandboxDirs(dirs...)` | File access outside the working directory |
| `WithSessionDir(dir)` | Where sessions persist |
| `WithInstruction(s)` / `WithExtraInstruction(s)` | Replace or append to the system prompt |
| `WithTools(...)` / `WithToolsets(...)` | Add your own |
| `WithBeforeToolCallbacks(...)` / `WithAfterToolCallbacks(...)` | Observe or rewrite tool calls |
| `WithBeforeModelCallbacks(...)` / `WithAfterModelCallbacks(...)` | Observe or rewrite LLM calls |
| `WithLSP(mode)` | `LSPOff`, `LSPMin` (default), `LSPFull` |
| `WithMemory(bool)` / `WithPalace(bool)` | Opt in to the shared `~/.pi-go` stores — **off by default** |
| `WithSkills(bool)` / `WithSubagents(bool)` | Toggle discovery — on by default |
| `WithAgentEvents(fn)` | Subagent and background-bash progress |

ADK types are used directly — no parallel type system. No `internal/` type
appears in an exported signature.

### What `WithModel` alone buys

Hooks, MCP servers, A2A agents and compactor settings from
`~/.pi-go/config.json` plus any project override; a sandbox rooted at the
working directory plus `~/.pi-go`; the core tool set with the bash supervisor
and control tools; skills from `.pi-go/skills/`, the user directory and the
bundled set; subagents from `.pi-go/agents` plus the bundled set; project rules
(`AGENT.md`/`AGENTS.md`/`CLAUDE.md`/`.pi-go/AGENTS.md`) discovered upward; MCP
and A2A toolsets; sessions under `~/.pi-go/sessions`, so `pi --resume` picks up
a session an embedder started.

## Callback composition, and an ADK detail not yet in the spec

ADK's after-tool chain has **two** defects, not one, both in
`invokeAfterToolCallbacks` (`adk/v2@v2.0.0/internal/llminternal/base_flow.go:1296`):

```go
for _, callback := range f.AfterToolCallbacks {
    result, err := callback(toolCtx, tool, fArgs, fResult, fErr)  // ← always fResult
    if err != nil {
        return nil, err
    }
    if result != nil {
        return result, nil                                        // ← short-circuits
    }
}
```

1. **It stops at the first callback returning a non-nil result.** This is the
   known one, and it is what `specs/memory-fixes/` documents.
2. **It passes the *original* `fResult` to every callback, not the previous
   one's output.** This is not in that spec, and it is arguably the worse half:
   even with the short-circuit fixed, a chain would not compose. pi-go's
   ordering assumes it does — dedup is registered after the compactor
   specifically so it compares calls in their post-compaction form, and under
   ADK's semantics it would compare the raw form instead.

pi-go registers several callbacks that all return the result map, so handing
ADK a slice runs only the first — silently disabling the compactor, dedup, the
LSP hook and memory recording.

`piagent` composes the whole after-tool chain — its own callbacks and the
embedder's — into a single ADK callback with explicit semantics:

- `(nil, nil)` — result unchanged, chain continues;
- `(m, nil)` — `m` becomes the result every later callback and the model sees;
- `(_, err)` — chain aborts, error surfaces.

Embedder callbacks run last, so they observe the compacted and deduplicated
result. Both properties are pinned by tests, including the feed-forward one.
Before-tool and model callbacks are passed as slices: ADK's "run until one
intervenes" is already right there.

This does not depend on the unmerged fix in `specs/memory-fixes/`.

## One deliberate difference from the CLI

**Memory and palace default off.** They are the only subsystems that *write* to
state shared with the user — `~/.pi-go/memory/claude-mem.db` and
`~/.pi-go/palace.db` are the same stores a real `pi` session uses, and an
embedder's process is not a `pi` session. Interleaving its observations with
the user's on construction is a surprise documentation cannot undo.
`WithMemory(true)` / `WithPalace(true)` opt in.

Skills and subagent discovery stay on: those *read* `.pi-go/`, which is the
reason to embed this agent rather than write one.

## Gates kept, not defaulted on

LSP registers only when a language server is installed, defaulting to `LSPMin`
(two tools of seven). Palace tools, once opted into, register only when the
palace holds at least one drawer. Tool declarations are billed on every
request.

## Finding the provider behind a model

The OTel `gen_ai.provider.name` attribute and the Gemini grounding tool both
need to know the provider. `piagent` asks the model — `m.(interface{ Provider() string })`, the *shape*,
never `pimodels.ProviderNamer`, so the dependency stays on ADK alone and the
isolation test still passes. Models from `pimodels` (#164) satisfy it.

Both branches are pinned end-to-end, not just on the helper: a model named
`internal-codename-7` reporting `Provider() == "gemini"` gets the grounding
tool, and a plain ADK model with no `Provider()` method still builds an agent
and completes a turn. The test double mirrors `pimodels.providerModel` exactly
— value receiver on a struct embedding `model.LLM` — because that detail is
load-bearing: `New` returns it by value, so a pointer-only method would leave
every consumer silently on the fallback path with no error anywhere.

## Two changes in `internal/agent`

- `Config.WorkingDir` — the directory reported to the model. Empty means the
  process's, so the CLI is unchanged; `piagent` sets it so an embedder driving
  a directory it was not started in is not told the wrong one. `CreateSession`'s
  default title uses the same value.
- `LoadInstructionPartsFor(base, cwd)` — exported wrapper over the existing
  `loadInstructionPartsFrom`, so project-rule and skill discovery follow the
  configured directory rather than `os.Getwd()`.

## Deliberately left out

- **Provider construction, in any form** — enforced by `isolation_test.go`.
- **The daily-token guardrail.** It wraps the model, not the agent.
- **The two-stage auto-compaction pre-turn hook.** `buildAutoCompactHook` is
  unexported in `internal/cli`; wiring it means extracting it first.
- **A process-global HTTP trace sink.** `httplog.SetSink` is global state.
- **The subagent orchestrator.** Subagents are reachable as tools; exposing the
  orchestrator is the larger commitment, so the conservative option shipped.

## Tests

`go test ./piagent/... -race` clean, coverage **87.4%** of statements. The
load-bearing tests are the end-to-end ones: an agent built with a fake
`model.LLM` that runs a real turn, and one where the fake emits a
`FunctionCall` so an `ls` tool call really executes and the embedder's
after-tool callback is proven to run on the post-compaction result.
`example_test.go` carries four compiled examples.

`make test`, `make vet`, `make lint` all clean.

## A pre-existing bug found on the way — #166

Not caused by this PR, and not fixed by it. Flagging it here so it is not
merged past as an implementation detail of the new package.

`initNonInteractiveRuntime` (`internal/cli/cli.go:515`) creates a subagent
event channel and **nothing ever reads it**:

```go
agentEventCh := make(chan tui.AgentSubEvent, 128)
agentEventCB := func(agentID, eventType, content string) {
	select {
	case agentEventCh <- tui.AgentSubEvent{...}:
	default:                       // silently dropped once full
	}
}
```

The channel is stored on `nonInteractiveRuntime` and never received from, so
the first 128 events fill the buffer and every event after that is discarded —
and the buffered 128 are never delivered either. The same callback is the sink
for backgrounded bash output (`cli.go:530`), so live output from background
commands goes nowhere in `--mode print`, `--mode json` and the RPC path. The
interactive path drains this channel, so the leak is specific to the
non-interactive runtime.

No output depends on these events today, so nothing is visibly broken — but it
is dead machinery that looks live, and a future feature wiring itself to
`agentEventCB` would appear to work and then stop after 128 events.

Tracked in **#166** with three options ranked by effort. `piagent` makes it a
caller's choice: `WithAgentEvents(fn)` delivers the events, and without one
they go to an explicit no-op rather than into a buffer nobody reads.


dimetron:
Superseded by #167, which combines this with #164 so the two public packages are reviewed as one API. All three commits from this branch are preserved there.

PR #181, opened 2026-08-21
feat(pimodels)!: take a defaultModel in New and Resolve

dimetron:
## What

`New` and `Resolve` both take a `defaultModel`, used when `modelName` is empty:

```go
func New(ctx context.Context, modelName, defaultModel string, opts ...Option) (Model, error)
func Resolve(modelName, defaultModel string, opts ...Option) (Info, error)
```

A model name read from a flag or a config file goes straight through, empty or
not, without the caller testing it for empty first:

```go
m, err := pimodels.New(ctx, *modelFlag, "claude-sonnet-5")
```

| `modelName` | `defaultModel` | Result |
|---|---|---|
| `""` | `"claude-sonnet-5"` | `anthropic` / `claude-sonnet-5` |
| `"gpt-5.6-luna"` | `"claude-sonnet-5"` | `openai` / `gpt-5.6-luna` — an explicit name always wins |
| `"cluade-sonet-5"` | `"claude-sonnet-5"` | error — a typo is not a missing name |
| `""` | `""` | error |

Both share one `applyDefault` helper, so they cannot drift into disagreeing
about what an empty name means.

## Why the asymmetry

The default covers a name nobody supplied, not one supplied wrongly. Falling
back on an unroutable name would send the request to a model the caller never
wrote, and nothing would say so — the mistake would surface in the output or the
bill, not at startup.

## Why `New` needs the default, not just `Resolve`

`Resolve`'s answer cannot be fed back into `New`. `Resolve` moves the routing
prefix out of the name and into `Info.Provider`, so `Info.Model` is no longer a
name that resolves:

```
claude-sonnet-5        -> anthropic claude-sonnet-5   ok
ollama/gemma4:e4b      -> ollama    gemma4:e4b        ROUND TRIP BROKEN
ollama/my-finetune     -> ollama    my-finetune       ROUND TRIP BROKEN
azure/prod-deployment  -> azure     prod-deployment   ROUND TRIP BROKEN
opencode/kimi-k3       -> opencode  kimi-k3           ROUND TRIP BROKEN
```

Every `ollama/`, `azure/` and `opencode/` name fails, because
`provider.OllamaModelPrefixes` is an empty slice
(`internal/provider/provider.go:174`) and nothing else reconstructs the routing.
Only the vendor-prefix names survive, which is why the hazard is easy to miss.

So before this change a caller holding a possibly-empty name had no correct
one-step option: `New` rejected the empty string, and resolving first to supply
a default corrupted the name on the way back. `New` taking the default itself
closes that gap.

The round-trip hazard itself is **not fixed here** — see the follow-up issue.
`TestNewNeedsNoResolveRoundTrip` pins current behaviour and skips itself if the
round trip is ever repaired.

## The both-empty guard

`New("", "")` and `Resolve("", "")` now error. That case was unreachable in
practice before; now a caller passes a possibly-empty flag on purpose. With
`WithBaseURL` set, `provider.ResolveWithBaseURL` treats an unrecognized name as
a custom OpenAI-compatible model and returns `Info{Provider: "openai",
Model: "", Custom: true}` — no error, empty model name. The explicit check turns
that into a real error.

## Breaking change

`New` and `Resolve` each gain a positional parameter. In-repo callers are
updated (`piagent/compose_test.go`, examples); external embedders on `v0.0.73`
or earlier pass `""` to keep the old behaviour. Worth a line in the release
notes.

## Verification

- `go build ./...`, `go vet`, `go test -count=1 ./pimodels/... ./piagent/...` pass
- `golangci-lint run ./pimodels/... ./piagent/...` — 0 issues
- Exercised from a **separate external module** (own `go.mod`, `replace` to this
  branch) driving a vanilla ADK `llmagent`:

| invocation | result |
|---|---|
| no flag | `provider: anthropic` — default applied |
| `-model gpt-5.6-luna` | `provider: openai` |
| `-model ollama/gemma4:e4b` | `provider: ollama` — the case the round trip broke |
| `-model cluade-sonet-5` | rejected, not defaulted |

https://claude.ai/code/session_018MsiGPng4RQ2Nt8viBj3wB

### DIFF
--- a/piagent/compose_test.go
+++ b/piagent/compose_test.go
@@ -26,7 +26,7 @@ func TestComposeModelIntoAgent(t *testing.T) {
 	// A local Ollama endpoint resolves and constructs without a credential and
 	// without reaching the network, which keeps this about composition rather
 	// than about a vendor being up.
-	m, err := pimodels.New(ctx, "ollama/gemma4:e4b",
+	m, err := pimodels.New(ctx, "ollama/gemma4:e4b", "",
 		pimodels.WithBaseURL("http://127.0.0.1:11434"))
 	if err != nil {
 		t.Fatalf("pimodels.New: %v", err)
@@ -58,7 +58,7 @@ func TestComposeModelIntoAgent(t *testing.T) {
 // that model can read it without importing pimodels. This is what let piagent
 // delete its own model-name prefix table.
 func TestComposeProviderNamerFlowsThrough(t *testing.T) {
-	m, err := pimodels.New(context.Background(), "ollama/gemma4:e4b",
+	m, err := pimodels.New(context.Background(), "ollama/gemma4:e4b", "",
 		pimodels.WithBaseURL("http://127.0.0.1:11434"))
 	if err != nil {
 		t.Fatalf("pimodels.New: %v", err)
--- a/pimodels/README.md
+++ b/pimodels/README.md
@@ -84,7 +84,7 @@ here that touches pi-go's configuration; `New` is self-contained.
 ## Inspecting before connecting
 
 ```go
-info, err := pimodels.Resolve("claude-sonnet-5")
+info, err := pimodels.Resolve("claude-sonnet-5", "")
 // info.Provider == "anthropic"
 
 pimodels.ContextWindow("gemini-3.7-flash")           // tokens, 0 if unknown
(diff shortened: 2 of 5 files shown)


dimetron:
Closing unmerged — abandoning the change. `pimodels.New` and `pimodels.Resolve` keep their existing signatures; `main` is unchanged.

The round-trip hazard this PR worked around is real and independent of it — tracked in #182.

PR #189, opened 2026-08-21
refactor: cut cyclomatic complexity across nine packages

dimetron:
A seven-agent pass over every function `gocyclo` scored above 25, following `.claude/skills/go-cyclomatic-complexity`.

| Metric | Before | After |
|---|---|---|
| Worst function | 37 | 29 (fenced) / 28 (deliberate keep) |
| Functions over 25 | 19 | **2** |
| Functions over 20 | 38 | **10** |
| Average | 4.04 | 3.98 |
| Coverage | 88.3% | **89.1%** |

Gates: `gofmt` clean · `go build ./...` · 38 packages pass · `golangci-lint run ./...` **0 issues** · `-race` clean.

## Worst offenders

| Function | Before | After | Pattern |
|---|---|---|---|
| `formatContextUsage` | 37 | **2** | section builders |
| `formatToolResult` | 34 | **3** | handler chain |
| `renderMessages` | 33 | **8** | role dispatch + sections |
| `ollamaContentsToMessages` | 31 | **8** | pure extraction + sections |
| `antContentsToMessages` | 31 | **8** | pure extraction + sections |
| `oaiContentsToMessages` | 30 | **7** | pure extraction + sections |
| `analyzeSessionFile` | 30 | **3** | named phases |
| `Orchestrator.Spawn` | 29 | **13** | setup + cleanup closure |
| `toolCallSummary` | 29 | **6** | lookup table |
| `deferredInit` | 28 | **10** | setup + cleanup closure |
| `Spawner.Spawn` | 28 | **4** | setup + cleanup, named phases |
| `smartTruncate` | 27 | **2** | pure classifier + sections |
| `modelPing` | 27 | **5** | named phases |
| `extractTriples` | 26 | **1** | section builders + pure helpers |
| `DrawerService.Search` | 25 | **7** | named phases + pure scorers |
| `aggregateTestOutput` | 22 | **4** | scan then render |

## How behavior preservation was verified

Not "the tests pass" — the tests were often the problem.

- **Characterization first.** Tests were written and run **green against the unmodified code** before any restructuring, so expectations describe observed behavior rather than the new implementation.
- **Byte-identical golden dumps of raw ANSI output** for every rendering function, captured before the refactor and diffed after — 66 frames for `View` across widths/scroll/matrix/loading states, 30 for `renderMessages` (each rendered twice to exercise the render cache), 39 tool-call and 45 result shapes.
- **20,000-input differential run** for `smartTruncate`, holding the original implementation beside the new one.
- **Mutation-tested the safety net itself.** One agent found its own goldens couldn't catch a forward/backward scan inversion (every case had a single candidate), so it checked the pre-refactor code back out of HEAD, regenerated all goldens from it, and re-ran. The mutation now fails 3 goldens + 2 unit tests.

## Deliberately not split

- **`handleSlashCommand` (28)** — 25 one-line command cases. A handler map needs an adapter closure per entry (three different signatures), turning `case "/compact": m.handleCompactCommand()` into table → closure → method. That hides the dispatch table.
- **`slashCommandDesc` (25)** — flat lookup, one string literal per command.
- **`View` (20 → 13)** — only the three separable pieces were extracted; the composition core interlocks `msgStart`/`msgRows`/`panelRows`/`inputCursorY` and feeds five memoized fields. Splitting means threading six values through helpers.
- **`buildRootRuntime` (29)** — fenced off; #188 rewrites that block.

## One intentional behavior change

`pumpACPSession` read `proc.err.Error()` **after** releasing the lock — an unsynchronized read of a field written under it. The string is now captured under the lock. Same output, one fewer data race.

## Thirteen pre-existing defects found — pinned, not fixed

Each is pinned by a test so a deliberate fix shows as a visible test change:

1. **`SpawnWithRetry` can never retry** — `awaitOutcome` reads agent status before `forwardEvents` writes it, so a crashed subagent always reads `"running"`. No production callers today, so latent rather than live.
2. **`retryStream` swallows a failing Ollama stream** — a 500 on `/api/chat` is retried ~7s then yields no error and no chunks, so `pi ping` reports a live model on an empty reply.
3. **`coerceValueAtKey` passes the parent's key into the child map** — nested objects inside array elements are never coerced.
4. `keywordSimilarity` returns scores >1 against a field documented 0–1.
5. `sqlite_store.go:369` truncates a float bm25 rank with `int(rank)` — why #4 rarely surfaces.
6. Search popup Down-wrap leaves the highlight off-screen (`selectNext` doesn't recompute scroll on wrap; `selectPrev` does).
7. Type-to-search drops all non-ASCII — `len(key.Text) == 1` counts bytes.
8. `smartTruncate` reports `applied=false` after dropping lines, so the caller discards the truncation.
9. `clipSummary` cuts by byte at 117, halving a straddling rune.
10. Failed-gate fence indents only the first line of multi-line output.
11. Ollama tool-call argument key order is map-iteration random — defeats byte-exact request caching.
12. Double coercion: the `.$` pass runs second and overwrites from the stale original.
13. `View` is not reproducible run-to-run (`matrixState.tick` seeds from `time.Now()`, plus a random spinner verb) — now documented, since it makes golden-frame tests of `View` impossible without pinning both.

**Two existing tests are named for paths they never exercise**: `TestSearch_SemanticPath` never calls `DrawerService.Search`, and `TestOllamaPingFullStreamingError` passes only because its mock also breaks the non-streaming call.

### DIFF
--- a/internal/cli/interactive.go
+++ b/internal/cli/interactive.go
@@ -141,138 +141,24 @@ func deferredInit(
 	res *initResources,
 ) {
 	initTotal := deferredInitTotal(cfg)
-	send := func(item string, done bool) {
+	send := initProgress(func(item string, done bool) {
 		ch <- tui.InitEvent{Item: item, Done: done, Total: initTotal}
-	}
+	})
 	fail := func(err error) {
 		ch <- tui.InitEvent{Err: err}
 	}
 
 	// --- Phase 1: Core tools (fast, needed by everything) ---
 	send("tools", false)
-
-	sandbox, err := tools.NewSandbox(sandboxRoot, worktreeDir)
+	coreTools, err := initCoreTools(sandboxRoot, worktreeDir, res)
 	if err != nil {
-		fail(fmt.Errorf("creating sandbox: %w", err))
+		fail(err)
 		return
 	}
-	res.sandbox = sandbox
-
-	// Allow agent tools to access ~/.pi-go/ (logs, sessions, config).
-	if home, hErr := os.UserHomeDir(); hErr == nil {
-		_ = sandbox.AddExtraDir(filepath.Join(home, ".pi-go"))
-	}
-
-	// The supervisor is built here, before the UI event channel exists, because
-	// the bash tool needs it at construction time. Its sink is attached later,
-	// once there is somewhere to stream to.
-	bashSup := tools.NewBashSupervisor()
-	res.bashSup = bashSup
-
-	coreTools, err := tools.CoreTools(sandbox, tools.WithBashSupervisor(bashSup))
-	if err != nil {
-		fail(fmt.Errorf("creating core tools: %w", err))
-		return
-	}
-	bashCtlTools, err := tools.BashControlTools(bashSup)
-	if err != nil {
-		fail(fmt.Errorf("creating bash control tools: %w", err))
-		return
-	}
-	coreTools = append(coreTools, bashCtlTools...)
-
 	send("tools", true)
 
 	// --- Phase 2: Parallel subsystems ---
-	type parallelState struct {
-		mu sync.Mutex
-
-		// Git + subagents
-		repoRoot     string
-		agentConfigs []subagent.AgentConfig
-		gitBranch    string
-		diffAdded    int
-		diffRemoved  int
-
-		// LSP
-		lspMgr   *lsp.Manager
-		lspTools []adktool.Tool
-
-		// MCP
-		mcpToolsets []adktool.Toolset
-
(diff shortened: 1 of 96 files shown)


dimetron:
Closing in favour of a fresh, evidence-backed refactor branch.

This branch grew to 96 files / +13,097 −2,915 in a single commit, which makes the behaviour-preservation argument hard to audit. Replacing it with a package-scoped series that carries measured before/after gocyclo + gocognit numbers and per-package coverage deltas for every function touched.

PR #190, opened 2026-08-21
fix: stop PDD /plan from losing spec artifacts in the worktree

dimetron:
## Summary

`/plan` writes spec files into an isolated git worktree and snapshots them via
`CommitAll → CreateBackupBranch → MergeBack → Cleanup`. Two bugs made that
snapshot silently drop every artifact, so `/run` found no `PROMPT.md`.

### Root cause

1. **`CommitAll` gated on `git status --porcelain`** (`internal/subagent/worktree.go`).
   Plain porcelain status hides ignored files, and the spec path matches the
   repo's ignore rules — `**/specs/` in the global gitignore and `.pi-go/` in the
   project — so status came back empty and `CommitAll` returned "nothing to
   commit" before staging anything. The backup branch then pointed at an empty
   commit, `MergeBack` merged nothing, and `Cleanup` ran `worktree remove
   --force`, deleting the only copy.
2. **`git add -A` respects ignore rules**, so even if staging was reached the
   ignored spec files were still not added.

### Fix

- **`CommitAll`**: force-add with `git add -Af` first, then check what was
  actually staged via `diff --cached --name-only` (preserves the empty-commit
  no-op). This guarantees ignored planner output enters the snapshot.
- **`finishPlanWorktree`**: copy the finished spec directory directly into the
  invoking checkout's `specs/` tree, so `/run` can read
  `<workDir>/specs/<task>/PROMPT.md` independently of any git merge.
- **Regression test**: `initTestRepoWithIgnores` installs the repo's ignore
  rules via `core.excludesFile`, so the loss is reproduced deterministically in
  any environment (the old test only failed on machines whose global gitignore
  happened to match).
- Added `copyDir` helper + unit tests in `internal/tui/plan_test.go`.

### Verification

- `go build ./...`
- `go vet ./internal/subagent/ ./internal/tui/`
- `go test ./internal/subagent/`
- `go test ./internal/tui/`


dimetron:
Closing: PR was opened against main but branched from feat/web-voice-gemini, which carries unrelated unmerged voice work. Rebasing the single fix commit onto main and reopening.

PR #207, opened 2026-08-22
feat(mcp): style OAuth callback page and cache tokens for 24h

dimetron:
## Summary
- Render the OAuth authorization-result page in the pi-go.sh visual style: dark neon theme (#0a0a12), grid + scanline overlays, Orbitron/Rajdhani type, glowing card with success (green ✓) / failure (magenta ✗) variants. All CSS is embedded so the page renders offline.
- Shared renderer `auth.RenderCallbackPage` now backs both callback paths: SSO PKCE `handleCallback` (success + error/state/no-code responses) and the MCP OAuth loopback callback.
- Persist MCP OAuth tokens per server under `~/.pi-go/mcp-oauth/` (0600) with a 24h reuse window (`mcpOAuthTokenTTL`).
- Reuse path: cached token is injected via the SDK's `InitialTokenSource`, so restarts within 24h skip the browser flow entirely; expired-without-refresh or stale entries are discarded.
- Capture path: freshly authorized tokens are persisted through `NewTokenSource`, storing `client_id`/`token_url` alongside the token so later sessions can rebuild a refreshing token source without re-running discovery/dynamic client registration.

## Follow-up (ffc88af) — codex review findings
- **Callback server now lives for the handler's lifetime.** The SDK re-runs `Authorize` (and so the code fetcher) on every 401 — e.g. a cached token that was revoked — and on `insufficient_scope` step-up, and the redirect URI registered via DCR carries the loopback port. Closing the listener after loading a cached token / after the first flow broke re-authorization. The fetcher also drops any stale outcome left from an earlier flow before opening the browser.
- **Provider error redirects (user clicked Deny) are surfaced to the fetcher** instead of leaving it waiting for the connect timeout: the callback delivers either the authorization result or the OAuth error on one result channel.
- Handler internals split into `mcpOAuthCallbackHandler`, `mcpOAuthCodeFetcher` (injectable URL opener) and `mcpOAuthNewTokenSource` so they are unit-testable without a listener or a browser.
- Dismissed (round 2, P2): "restore cached granted scopes before step-up". `auth.AuthorizationCodeHandlerConfig` has no way to seed the SDK's private `grantedScopes` map; the only alternative is not caching at all. Worst case after a restart is one extra authorization prompt on an `insufficient_scope` step-up, which is the pre-PR behaviour on every start. Would need an upstream SDK hook.

## Test plan
- `go build ./...`, `go vet`, `make lint` (0 issues), `go test -race ./internal/extension/`
- `go test ./internal/auth/ ./internal/extension/ -count=1` — new tests cover: page rendering/escaping/failure variant; callback handler (error / no-code / success / duplicate); fetcher (result, stale drain, provider error, ctx cancel); `NewTokenSource` hook; `InitialTokenSource` reuse of a cached token; store branches (identity mismatch inside the file, corrupt JSON, expired-without-refresh, static fallback, real refresh against a fake token endpoint with rotated-token write-back, mkdir/rename/home errors, permissions). Store tests set `USERPROFILE` as well as `HOME` so they are hermetic on Windows. `internal/extension` coverage 85.2% → 91.3%.

## CI notes
- The earlier **Lint** failure was `golangci-lint config verify` timing out while downloading the JSON schema — infra, not a finding. Green on re-run.
- **Test (Windows)** fails on `main` too (since `cd33d6a`), with the same set of pre-existing failures (`internal/auth` `TestSaveKey*`, `internal/browser`, `internal/acp/*`, `internal/cli` 600s timeout, `internal/extension` `hooks_test.go`/`complexity_extension_test.go`, …). None of this PR's tests fail on Windows. PR #203 tracks the Windows fix.
- One `Test` (Linux) run hit a pre-existing data race in `internal/acp/client/test_helper_agent_test.go` (`agentConn` assigned after `NewAgentSideConnection` starts receiving); passed on re-run. Not touched by this PR.

https://claude.ai/code/session_014XA8pYe34AXYE8sSeRe3WC

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -729,7 +729,7 @@ func handleCallback(w http.ResponseWriter, r *http.Request, expectedState string
 		}
 		logf("callback: oauth error=%s description=%q", errParam, desc)
 		ch <- codeResult{err: fmt.Errorf("OAuth error: %s", desc)}
-		http.Error(w, "Authentication failed: "+desc, http.StatusBadRequest)
+		RenderCallbackPage(w, http.StatusBadRequest, false, "Authentication failed", desc)
 		return
 	}
 
@@ -744,19 +744,14 @@ func handleCallback(w http.ResponseWriter, r *http.Request, expectedState string
 	if code == "" {
 		logf("callback: no code received")
 		ch <- codeResult{err: fmt.Errorf("no authorization code received")}
-		http.Error(w, "No code received", http.StatusBadRequest)
+		RenderCallbackPage(w, http.StatusBadRequest, false, "No code received", "")
 		return
 	}
 
 	logf("callback: ok code_len=%d", len(code))
 	ch <- codeResult{code: code}
 
-	w.Header().Set("Content-Type", "text/html")
-	_, _ = fmt.Fprint(w, `<!DOCTYPE html><html><body>
-<h2>✓ Authentication successful</h2>
-<p>You can close this tab and return to pi.</p>
-<script>window.close()</script>
-</body></html>`)
+	RenderCallbackPage(w, http.StatusOK, true, "Authentication successful", "You can close this tab and return to pi.")
 }
 
 func exchangeCode(ctx context.Context, prov Provider, code, redirectURI, verifier string) (*TokenResponse, error) {
--- a/internal/auth/callback_page.go
+++ b/internal/auth/callback_page.go
@@ -1,0 +1,184 @@
+package auth
+
+import (
+	"html"
+	"io"
+	"net/http"
+	"strings"
+)
+
+// callbackPageTemplate renders the OAuth authorization-result page shown in
+// the user's browser after the identity provider redirects back to the local
+// loopback callback server. Visual language matches https://pi-go.sh/ (dark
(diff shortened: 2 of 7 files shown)


dimetron:
Codex review findings addressed in 7fcfb75:

- **P1 (token bound only to name)** — cache entries now keyed by name + MCP URL (full SHA-256 of both), and loading validates the stored identity; a retargeted/renamed server can no longer replay another endpoint's token.
- **P1 (rotated refresh tokens never persisted)** — new `persistingTokenSource` wraps both the cached-token source and the freshly-authorized source, saving whenever access/refresh credentials change.
- **P2 (perms not enforced on existing paths)** — directory is chmod'd to 0700 and the file is written atomically via `os.CreateTemp` (0600) + rename.
- **P2 (listener leak / unguarded sends)** — the loopback server is closed when a cached token makes it unnecessary, the fetcher closes it after the first result, and callback delivery is a non-blocking one-shot send that ignores duplicates.
- **P2 (error redirects rendered as success)** — the MCP callback now validates `error`/missing `code` first and renders the failure page.

New tests: identity-drift rejection (URL + name), rotated-token persistence, filename collision across URLs. `go build ./...`, `go vet`, and both package test suites pass.

PR #241, opened 2026-08-27
feat(provider): mistral provider routing fix + PDD spec TOO/024 (all commits signed)

dimetron:
## Summary

Re-lands the 25 local commits that were reset from `main` (backup branch `backup/reset-to-origin-main-20260827`), now **all cryptographically signed** with `Signed-off-by` trailers, **plus** the missing model-list work from the spec.

- **27 commits** on this branch (3 workflow merge commits and 1 empty duplicate dropped by the linear rebase)
- Every commit verified: `gpgsig` header present + matching `Signed-off-by` trailer

## What's in the branch

1. **PDD spec `features/TOO/024-mistral-provider`** — full spec: rough-idea, requirements, research (routing-bug, provider-layer, model-catalog, reference-ts, build-test-e2e), design, plan (9 slices), PROMPT.
2. **Slice 1** (`9334d5a`): `mistral/` prefix auto-detect + strip in `provider.Resolve`, `mistral`/`magistral` prefixes in config `autoDetectProvider`, unit tests.
3. **Slices 5–8** (`81329a2`, `7d879e5`, `1a8eb3c`): model-list enrichment, `-o json`, refreshable catalog, fetched catalogs.
4. **All six provider catalogs + a guard** (`73ec9c5`): see below.
5. **Windows cache isolation + coverage** (`bc7f045`): see below.
6. **Codex review fixes** (`64e44b4`): see below.
7. **Hook fix** (`9a297de`): scope pre-commit golangci-lint to repo packages.

## All providers present in `modeldata/` (`73ec9c5`)

Four of the six providers had a checked-in `models-<provider>.json`; `anthropic` and `openai` did not. `make fetch-models` had not failed — it **skipped both silently**.

The skip was in the script's key check, not the binary. `fetch-models.sh` sourced only `$HOME/.pi-go/.env`, while the CLI's `loadDotEnv` reads that file *and then* the nearest `.pi-go/.env` walking up from the working directory. In this repo the second file holds `ANTHROPIC_API_KEY`, so `pi model list anthropic` would have worked while the script decided the key was missing. The script now parses both files the way `loadDotEnvFile` does (`KEY=VALUE`, blank/`#` skipped, empty values ignored — parsed, not executed) and walks up for the project file.

All six regenerated. The four existing files came back byte-identical, confirming the fetch normalization is idempotent:

| Provider | Models | Status |
|---|---|---|
| `anthropic` | 10 | **new** |
| `openai` | 18 | **new** |
| `gemini` | 50 | unchanged |
| `mistral` | 37 | unchanged |
| `xai` | 12 | unchanged |
| `openrouter` | 417 | unchanged |

`ollama` is excluded by design — its model list is whatever the local daemon has pulled, so a checked-in snapshot would describe one developer's machine. `ValidateModel` exempts Ollama for the same reason.

Three tests keep the next silent skip from shipping:

- `TestEmbeddedCatalogPresentForEveryProvider` — each snapshot present, valid JSON, non-empty, deduped, sorted by id.
- `TestEmbeddedCatalogFeedsCatalogFor` — each one actually reaches `CatalogFor`. `openrouter` is load-bearing: it has no hard-coded `KnownModels` entry, so without its snapshot `CatalogFor` returns nothing and `ValidateModel` stops validating OpenRouter models at all.
- `TestFetchModelsScriptCoversEveryProvider` — pins `fetch-models.sh`'s `PROVIDERS` list against the same set, so adding a provider to one and not the other fails.

### Two `TestResolveRuntimeModel` fixes that fall out of this

- The **"role provider name overrides the resolved provider"** case paired provider `openrouter` with the bare Anthropic ID `claude-sonnet-4-6`. That pair is invalid — OpenRouter names the model `anthropic/claude-sonnet-4.6` — and was only passing because `CatalogFor("openrouter")` was empty and `ValidateModel` skipped the provider. Split into two cases: the override itself is asserted through `opencode` (no catalog), and a second case pins the rejection.
- Added `isolateModelCatalog` to the table. `CatalogFor` prefers the XDG cache over the embedded snapshot, and `ValidateModel` refreshes over the network when a provider key is exported — so the table validated against whatever the developer's machine had cached, and spent **26s in live HTTP** on a miss. Now hermetic.

## Windows cache isolation + coverage (`bc7f045`)

Two CI failures, both pre-dating the catalog work and surfaced by it.

**Windows.** `withTempCacheDir` set `HOME` and `XDG_CACHE_HOME`, but `os.UserCacheDir` reads exactly one variable per platform and `%LocalAppData%` is the Windows one — so the helper was a no-op there. `TestRefreshCatalogWritesCache` wrote its one-model catalog into the runner's real cache, and every later `CatalogFor("mistral")` in the package read that instead of the embedded snapshot. That failed `TestValidateModel/mistral/{codestral,pixtral,ministral}` and, once it existed, `TestEmbeddedCatalogFeedsCatalogFor`. All three variables are now set, here and in the cli-side `isolateModelCatalog`.

**Coverage.** `codecov/patch` was at 85.81% against a 91.03% target. Covered the lines slices 5–8 added and left untested: `RefreshCatalog`'s write-side error paths (no resolvable cache dir, `MkdirAll` failure, `WriteFile` failure — the two error cases pin that the fetched models come back *alongside* the error), `modelsCacheDir` when `os.UserCacheDir` fails, `apiKeyForProvider` across all six providers plus the unlisted-provider `""`, `listMistralModels`' full six-capability mapping and its `/v1` base-URL branch, and `-o json` sorting by id.

Uncovered added lines: **35 → 13**. The remainder are unreachable from a unit test — three `json.Marshal` error branches on structs that cannot fail to encode, the `os.Rename` failure after a successful write, the embedded-catalog unmarshal failure that `TestEmbeddedCatalogPresentForEveryProvider` now rules out, and `ValidateModel`'s refresh-on-miss, which builds `ListModelsOptions` with no `BaseURL` and so has no seam to point at a test server.

## Codex review round (`64e44b4`)

Codex posted [review #5040656424](https://github.com/dimetron/pi-go/pull/241#pullrequestreview-5040656424) with six findings. Each was verified against the code before being accepted or dismissed; four were real defects, one was a documentation error, one was out of scope.

**`ValidateModel` discarded the models it had just fetched.** The refresh-on-miss path called `RefreshCatalog`, threw the returned slice away, and re-read `CatalogFor`. But `RefreshCatalog` deliberately returns the fetched models even when it cannot persist them, and the re-read then sees only the embedded snapshot — so a model the provider had just confirmed was rejected because *caching* failed. Now matched against the returned slice.

**The refresh ignored the configured endpoint.** It built `ListModelsOptions` with no `BaseURL`, so a user pointed at a gateway got a refresh against the vendor's public API — 401, and a valid model reported as unknown. Added `baseURLForProvider`, reading the same env vars `config.BaseURLs` does. This also gave the path a test seam, so it is now covered.

**`TestValidateModel_RefreshOnMiss` proved nothing.** Its model `codestral-2508` matches the hard-coded `codestral` prefix and is in the embedded snapshot, so `ValidateModel` returned before the refresh ran — and the `MISTRAL_BASE_URL` it set was never read anyway. Rewritten around a sentinel ID in neither catalog, asserting the provider was actually contacted, with a guard that fails if the sentinel ever becomes real. Two cases added: the no-usable-cache path, and one pinning that the refresh is not a blanket pass. The first fails against the old logic.

**`fetch-models.sh` destroyed the catalogs it exists to maintain.** Shell redirection truncated `models-$p.json` before the CLI started, and the failure branch deleted it — so a transient outage wiped the committed fallback. Now staged in a scratch dir and moved over the destination only on full success; an empty model list counts as a failure. **This paid for itself immediately**: the first run after the change hit a network blip on all six providers and left every catalog byte-identical, where the old script would have deleted all six.

**`RefreshCatalog` shared one temp filename.** Two concurrent validation misses for a provider both wrote `<path>.tmp`. Now `os.CreateTemp` in the target directory with deferred cleanup.

*Dismissed*: the request to land slices 2–4 and 9 — deliberately out of scope, and already recorded as such in the slice table below. *Documentation only*: the "one JSON document per line" claim was wrong in this description (the flag's own help says "one JSON document per provider", which `MarshalIndent` satisfies); corrected above rather than un-indenting output that has to stay reviewable as a checked-in diff.

All six threads answered and resolved.

## Spec validation (features/TOO/024-mistral-provider)

| Slice | Status |
|-------|--------|
| 1. Routing fix (`mistral/` strip + config auto-detect) | ✅ Implemented + tested |
| 2. NewMistral signature + thinkingLevel | ✅ Implemented + tested |
| 3. Reasoning/cache injection | ✅ Implemented + tested |
| 4. Thinking extraction | ✅ Implemented + tested |
| 5. Model list + `-o json` | ✅ Implemented + tested |
| 6. Catalog manager | ✅ Implemented + tested |
| 7. Wire catalog into ValidateModel | ✅ Implemented + tested |
| 8. Makefile fetch-models + embedded JSON | ✅ Implemented + all six providers committed + guarded by tests |
| 9. E2E tests | ✅ Implemented (gated on `MISTRAL_API_KEY`) |

**All nine slices are now implemented** (`5faffcb` lands 2–4 and 9). The spec's `summary.md` phase is not yet written.

### Known gap, not fixed here

A role cannot express "OpenRouter + `anthropic/claude-sonnet-4.6`" through the provider override alone: `provider.Resolve` runs first and only strips `openrouter/`, `ollama/`, `azure/`, `opencode/` and `mistral/` prefixes, so a vendor-qualified OpenRouter ID fails to resolve before the override is applied. Configuring a base URL for the role provider works around it. Changing `Resolve`'s prefix semantics is out of scope for this PR.

## What Slices 5–8 add

- `ModelInfo` gains `ContextWindow`/`Capabilities` (JSON tags `context_window`/`capabilities`)
- Dedicated `listMistralModels` card parser: parses `max_context_length` + `capabilities`, filters to `completion_chat`-capable models
- `pi model list mistral` shows context window + capabilities; other providers' output unchanged
- New `-o json` / `--output json` flag: one pretty-printed JSON document per provider (`{"provider","fetched_at","models":[...]}`), concatenated when several providers are listed — a stream, not JSONL. It is indented because `make fetch-models` writes it straight into `modeldata/`, where a one-line-per-catalog artifact would be unreviewable in a diff
- `catalog.go`: `CatalogFor`/`RefreshCatalog` with XDG cache (`os.UserCacheDir()/pi-go/models/<provider>.json`), atomic writes
- `ValidateModel` consults the runtime catalog and refreshes once on miss when an API key is present
- `make fetch-models` + `scripts/fetch-models.sh` shelling out to `pi model list <provider> -o json`

## What Slices 2–4 and 9 add (`5faffcb`)

`NewMistral` now takes a `thinkingLevel` (mirroring `NewOpenRouter`) and `NewLLM` passes it through. Each instance gets one `prompt_cache_key` (uuid), so every turn of a session shares the cached prefix — the xAI precedent.

Three points where the implementation deviates from `plan.md`, each recorded in the spec with its reason:

**One `SetExtraFields` call, not two.** `param.metadata.SetExtraFields` *replaces* the extra-fields map rather than merging into it (openai-go v3.52.0, `packages/param/param.go:157`). The plan's two consecutive calls would have dropped `prompt_cache_key` on every reasoning request. `mistralExtraFields` builds one map and sets it once.

**`reasoning_effort` has two values, not five.** The plan mapped `low→low`, `medium→medium`, `max→xhigh`. Mistral's [reasoning docs](https://docs.mistral.ai/studio-api/conversations/reasoning) document exactly `"high"` and `"none"`, and the reference TS agrees (`mapReasoningEffort` collapses every active level onto `"high"`). Sending the other three would put undefined values on the wire. Implemented: `none`/`off` → `"none"`, every active level → `"high"`, empty or unrecognized → field omitted. `none → "none"` is kept rather than omitting the field — pi's default level is `high`, so a user asking for `none` wants thinking off, and only the explicit value does that. `prompt_mode` has no off value, so on the magistral path the field is omitted instead.

**Thinking extraction needed two hooks, not one.** The plan assumed the existing runner "already yields the text parts separately". It does not. Mistral puts thinking *inside* `content`, which becomes an array of typed chunks for the thinking phase. openai-go declares `Content` as a `string` and its decoder does not fail on the array: `newPrimitiveTypeDecoder` calls `v.SetString(n.String())` *before* reporting the type mismatch, and the struct decoder swallows per-field errors (`internal/apijson/decoder.go:514`, `:475`). The raw array JSON is therefore left sitting in `Delta.Content` / `Message.Content`, and the existing runners would have emitted it verbatim as the assistant's answer — raw JSON in the transcript for exactly the turns thinking was enabled for. Verified against the SDK, not inferred.

So the shared runners now take a hooks struct:

```go
type oaiExtractHooks struct {
    deltaThinking   func(rawChunk string) string
    messageThinking func(rawResponse string) string
    answerText      func(content string) string
}
```

`oaiRunStreamingHooks` / `oaiRunNonStreamingHooks` take it; `oaiRunStreamingExtract` / `oaiRunNonStreamingExtract` remain as thin wrappers, so the OpenRouter call sites are untouched. `mistralAnswerText` reduces a chunk array to its `"text"` chunks and passes anything else through unchanged — an answer that merely looks like JSON (`[see note 1]`, `["a","b"]`, an array of chunk types this code does not know) arrives intact.

Tests: request-body capture across the model × level matrix (9 cases), `prompt_cache_key` stability within an instance and uniqueness across instances, the effort and model-gating tables, both extraction hooks including malformed input, and full streaming and non-streaming runs asserting thinking partials arrive before an intact answer. Four gated e2e tests, one of which (`TestE2EMistralReasoningEffortAccepted`) sends both documented `reasoning_effort` values against the live API so the mapping cannot drift unnoticed.

## Gates

- `go build ./...` ✅
- `go vet ./...` ✅
- `make test-unit` ✅ (full suite, run outside the sandbox — the `httptest.NewServer` trap is documented in CLAUDE.md)
- CI: all 12 checks green, including Test (Windows), on the final commit
- `make fetch-models` verified end-to-end against a stub: success writes sorted + deduped output; a failing provider and an empty model list both leave the existing catalog untouched
- `codecov/patch` 93.61% (target 91.03%) ✅ · `codecov/project` 91.04% (+0.01%) ✅
- `golangci-lint run ./cmd/... ./internal/... ./hack/...` ✅ (0 issues)
- `golangci-lint run ./cmd/... ./internal/... ./hack/...` ✅ (0 issues) on `5faffcb`
- `go test -tags e2e ./internal/provider -run Mistral` ✅ — **all 10 run live against api.mistral.ai with a real key**, not just compiled. The reasoning tests exercised the path they exist for: `TestE2EMistralReasoningStreaming` received 1425 characters of thinking as `thinking`-role partials ahead of an intact answer, and `TestE2EMistralNonStreamingThinking` got the thinking back as its own part. Both assert the answer does not begin with `[{`, which is what a regression on the array-content handling would look like. `TestE2EMistralReasoningEffortAccepted` confirmed the live API accepts both `"high"` and `"none"`
- All 28 commits signed + `Signed-off-by` ✅


specs/features/TOO/024-mistral-provider/requirements.md
```diff
@@ -0,0 +1,218 @@
...
+  opening the PR.
+
+### Requirements summary (consolidated)
+
+**Scope:**
+1. Fix `unknown openai model "mistral/codestral-2508"` — the validation/routing bug.
+2. `mistral/` prefix = auto-detect + strip (like azure/, ollama/).
+3. `internal/config` `autoDetectProvider` gains `mistral`/`magistral` prefixes.
+4. Dedicated Mistral model-list parse/print with more card fields (Q5b).
+5. Mistral chat: `reasoning_effort`/`prompt_mode` from thinking level +
```
dimetron, line 207: [P1] Implement the required Mistral chat slice before treating this feature as complete. The diff never changes `mistral.go`: `NewMistral` still receives no thinking level, and there is no Mistral-side `reasoning_effort`, `prompt_mode`, or per-instance `prompt_cache_key` injection anywhere in the repository. The corresponding new chat unit/E2E coverage is also absent. Either land slices 2–4 and 9 described by this PR's own spec, or narrow the PR/spec so it does not claim this behavior.

internal/provider/provider.go
```diff
@@ -302,20 +302,58 @@ func ValidateModel(info Info) error {
...
-	for _, prefix := range known {
-		if strings.HasPrefix(lower, prefix) {
-			return nil
+	if matchPrefix(known, lower) {
+		return nil
+	}
+	// Validation miss: refresh once when an API key is available, then
+	// re-check against the fresh catalog. Network errors are non-fatal.
+	if key := apiKeyForProvider(info.Provider); key != "" {
+		if _, err := RefreshCatalog(context.Background(), info.Provider, ListModelsOptions{APIKey: key}); err == nil {
```
dimetron: [P1] Validate against the fetched models even when they cannot be persisted. `RefreshCatalog` intentionally returns the live `models` when caching is disabled and returns `models` plus an error on write/rename failure, but this code discards that slice and only re-reads `CatalogFor` after a nil error. On a machine without a usable cache directory (or with a read-only/full cache), a model confirmed by the provider is still rejected. Capture the returned list and match it directly; cache persistence should not determine validation correctness.

scripts/fetch-models.sh
```diff
@@ -0,0 +1,104 @@
...
+    openrouter) KEY_VAR="OPENROUTER_API_KEY" ;;
+  esac
+
+  if [ -z "${!KEY_VAR:-}" ]; then
+    echo "skip $p: $KEY_VAR not set"
+    continue
+  fi
+
+  echo "fetching $p..."
+  if ! /tmp/pi-fetch-models model list "$p" -o json > "$OUT_DIR/models-$p.json" 2> /tmp/pi-fetch-models-$p.err; then
```
dimetron: [P1] Preserve the last-known-good embedded catalog on fetch failure. Shell redirection truncates `models-$p.json` before the command starts, and the failure branch then removes it, so a transient outage/403 deletes the committed fallback that this feature is meant to retain. Write the CLI output to a unique temporary file, validate/normalize it, and rename over the destination only after the whole fetch succeeds.

internal/provider/provider_test.go
```diff
@@ -271,6 +313,80 @@ func TestValidateModel(t *testing.T) {
...
+					},
+				},
+			},
+		})
+	}))
+	defer srv.Close()
+	t.Setenv("MISTRAL_API_KEY", "testkey")
+	t.Setenv("MISTRAL_BASE_URL", srv.URL)
+
+	if err := ValidateModel(Info{Provider: "mistral", Model: "codestral-2508"}); err != nil {
```
dimetron: [P1] This does not exercise refresh-on-miss. `codestral-2508` is now in the embedded JSON and also matches the hard-coded `codestral` prefix, so `ValidateModel` returns before contacting `srv`; the test passes even though the implementation does not use the `MISTRAL_BASE_URL` set above. Use a sentinel ID absent from both embedded and hard-coded catalogs and assert that the handler was called (and, ideally, that the refreshed cache was used).

dimetron (commented):
Review of `git diff main...HEAD`. I found six actionable issues: the committed requirements include Mistral chat behavior that is not implemented, and the new catalog/output paths have correctness and atomicity gaps. Targeted tests (`go test ./internal/provider ./internal/cli ./internal/config`) pass, but one refresh test is a false positive as noted inline.

— Codex review

internal/provider/catalog.go
```diff
@@ -0,0 +1,129 @@
...
+		Provider:  providerName,
+		FetchedAt: time.Now().UTC().Format(time.RFC3339),
+		Models:    models,
+	}
+	b, err := json.MarshalIndent(cf, "", "  ")
+	if err != nil {
+		return models, fmt.Errorf("encoding catalog: %w", err)
+	}
+	path := cachePath(providerName)
+	tmp := path + ".tmp"
```
dimetron: [P2] Use a unique temp file for the atomic write. Concurrent validation misses for the same provider share `<path>.tmp`; their writes/renames can race, causing one refresh to install another caller's bytes or fail after the other rename, and rename failures leave the shared temp behind. Follow the existing `os.CreateTemp`-in-target-directory pattern, close/fsync as appropriate, defer cleanup, then rename.

internal/cli/model.go
```diff
@@ -99,6 +111,27 @@ func runModelList(cmd *cobra.Command, args []string) error {
...
+				Provider:  p,
+				FetchedAt: time.Now().UTC().Format(time.RFC3339),
+				Models:    models,
+			}
+			// Sort by ID and pretty-print so the checked-in snapshots are
+			// stable and diff-friendly across fetches.
+			sort.Slice(doc.Models, func(i, j int) bool {
+				return doc.Models[i].ID < doc.Models[j].ID
+			})
+			b, err := json.MarshalIndent(doc, "", "  ")
```
dimetron, line 125: [P2] This contradicts the JSONL contract used below: `MarshalIndent` emits each provider document across many lines, so `pi model list -o json` with multiple providers is not one JSON document per line and line-oriented consumers cannot parse it as JSONL. Emit compact JSON (for example with an unindented `json.Encoder`) for the CLI stream; the fetch script can pretty-print the single-provider artifact afterward.

internal/provider/provider.go
dimetron: Accepted, fixed in `64e44b4`.

Confirmed: `RefreshCatalog` returns the fetched models even when it cannot persist them (no resolvable cache dir, read-only/full cache) and also returns them alongside a write error — and `ValidateModel` threw that slice away, then re-read `CatalogFor`, which at that point sees only the embedded snapshot. A model the provider had just confirmed was rejected because caching failed.

Now matched against the returned slice, and a non-empty result is used whether or not the write succeeded.

While verifying this I found the related half: the refresh built `ListModelsOptions` with no `BaseURL`, so a user pointed at a gateway got a refresh against the vendor's public API (401 → "unknown model"). Added `baseURLForProvider`, reading the same env vars `config.BaseURLs` does. That also gave this path a test seam, so it is now covered.

New: `TestValidateModel_RefreshOnMissWithoutCache`, which fails against the old logic and passes with the fix.

scripts/fetch-models.sh
dimetron: Accepted, fixed in `64e44b4`.

Confirmed and worse than theoretical. The fetch, normalize and validate steps now run in a `mktemp -d` scratch dir, and the result is moved over `models-$p.json` only after all of them succeed; an empty model list counts as a failure rather than a replacement, and the binary itself was moved out of `/tmp` into the same scratch dir with an `EXIT` trap.

This paid for itself immediately: the first `make fetch-models` run after the change hit a network blip and failed on all six providers, printing `FAILED (keeping existing ...)` for each and leaving every committed catalog byte-identical. Under the old script that run would have deleted all six.

internal/provider/provider_test.go
dimetron: Accepted, fixed in `64e44b4`.

Confirmed on both counts: `codestral-2508` matches the hard-coded `codestral` prefix *and* is in the embedded snapshot, so `ValidateModel` returned at the first `matchPrefix` — and the `MISTRAL_BASE_URL` the test set was never read by the implementation anyway.

Rewritten around a sentinel ID in neither catalog, and the handler now records whether it was called so the test fails if the refresh never runs. Added a guard that fails loudly if the sentinel ever becomes a real model. Two more cases: the no-usable-cache path from the `provider.go:316` thread, and one pinning that the refresh does not become a blanket pass for a model the provider does not list.

`MISTRAL_BASE_URL` is now genuinely honoured — see the `provider.go:316` thread.

internal/provider/catalog.go
dimetron: Accepted, fixed in `64e44b4`.

Now `os.CreateTemp(dir, providerName+".*.json.tmp")` in the target directory, with a deferred `os.Remove` that is a no-op once the rename succeeds, and an explicit `Chmod` to 0644 since `CreateTemp` creates at 0600.

internal/cli/model.go
dimetron: Half accepted — the documentation was wrong, not the code.

Checked the tracked contract: the flag's own help string is `"json" emits one JSON document per provider`, with no per-line claim. `MarshalIndent` satisfies that. The JSONL wording came from the PR description, which I have corrected.

Keeping the indentation deliberately: `make fetch-models` pipes this output straight into `internal/provider/modeldata/models-<provider>.json`, and those files are reviewed as diffs — a 417-model OpenRouter catalog on one line is unreviewable, and every fetch would show as a single changed line. That reviewability is the point of the normalization step (sort by id, dedupe, pinned `fetched_at`) right next to it.

On the multi-provider stream: concatenated pretty documents are still valid input for a streaming `json.Decoder`, and `fetch-models.sh` only ever requests one provider at a time, so nothing in the repo consumes it line-wise. If a line-oriented consumer shows up, the right shape is a `--output jsonl` alongside `--output json` rather than making the checked-in catalogs unreadable.

specs/features/TOO/024-mistral-provider/requirements.md
dimetron: Dismissed — out of scope by design, not an oversight.

Slices 2-4 (`NewMistral` thinking level, `reasoning_effort`/`prompt_mode`/`prompt_cache_key` injection, thinking extraction) and slice 9 (e2e) are deliberately not in this PR. The spec's own slice table in the PR description marks each of them `Not implemented`, and the description states they remain for a follow-up, along with the spec's unwritten `summary.md` phase.

This PR is the routing fix (slice 1) plus the model-catalog work (slices 5-8). `mistral.go` is untouched on purpose — the finding correctly observes that, which is what the slice table already says.

The spec is the plan for the feature; it is expected to describe more than any one PR lands. Narrowing it to match this PR would lose the plan for the remaining work.

PR #246, opened 2026-08-28
fix(agent): a toolset tool sharing a built-in's name no longer fails the turn

dimetron:
## The bug

Adding the `claude-chrome` MCP server to `~/.pi-go/mcp.json` broke **every** request with:

```
duplicate tool: "find"
```

The server exposes a `find` tool; pi-go has a built-in `find` tool. The ADK packs
the agent's own tools first and each toolset's tools after, and the packer rejects
a second tool claiming a name that is already taken — which fails the whole turn,
not just the offending tool. Reproduced in session `260828-1255-32384-b5981`; the
only workaround was removing the server from `mcp.json`.

## Why nothing caught it

Nothing was making tool names unique across sources. `internal/extension` had a
`deduplicateTools` hook, but it returned its input unchanged, and it could only
ever see one server's tools anyway — the doc comment above `resilientToolset`
claimed a dedup that was not happening.

## The fix

`buildRunner` is the one place that sees the built-in tools and every toolset at
once, so `dedupeToolsets` goes there.

- **Built-ins keep their names.** The system prompt and the tool-display code name
  them explicitly, so they win every collision.
- **A colliding toolset tool is renamed to `<toolset>_<tool>`** and stays callable.
  Only the model-facing name changes — `Run` delegates to the inner tool, which
  still addresses its MCP server under the tool's original name.
- **Assignments are remembered per toolset.** `Tools()` runs once per request; a
  name that drifted between turns would orphan a pending function call.
- Toolset-vs-toolset collisions are handled by the same shared claim set, and
  names are sanitized and capped at the 64-character API limit.

Users see a notice when a rename happens, routed through `internal/notice` (never
stdout — the TUI owns the terminal).

## Also in here

A stale e2e assertion expected `tools.LSPTools` to return all seven LSP tools. It
has defaulted to the minimal pair since the `LSPMin`/`LSPFull` split, so
`go test -tags e2e ./...` was **already red on main**. The test now covers both
modes, which makes the e2e gate meaningful again.

## Verification

| Gate | Result |
|---|---|
| `go test ./...` | pass |
| `go test -tags integration ./...` | pass |
| `go test -tags e2e ./...` | pass (was red on main — see above) |
| `make lint` | 0 issues |
| `go vet ./...` | clean |
| Coverage, `internal/agent/toolnames.go` | 100% of statements, all 14 functions |
| Coverage, repo total | 91.7% |

The new e2e test `TestE2EToolsetNameCollisionKeepsBothCallable` drives a real
agent turn with a toolset whose tool collides with the built-in `find`. It asserts
both tools are declared, both are callable, and the MCP tool actually ran. Removing
the `dedupeToolsets` call makes it fail with the exact production error:

```
Run() error: duplicate tool: "find"
```

Note: tests that bind a local listener fail under the agent sandbox
(`httptest.NewServer` panics in `newLocalListener`). All suites above were run
outside it.


dimetron:
### The TUI, after the fix

`./pi` on this branch, prompt `show me your tools`. The sidebar shows **MCP Tools [29]** resolving across three servers and the turn completing — the exact path that used to die with `duplicate tool: "find"`.

pi-go TUI answering "show me your tools"

### e2e: agent tool calls, including the name-collision regression test

`go test -count=1 -tags e2e ./internal/agent/ -run TestE2E -v` — 19 tests, all green.

e2e run

<sub>Recorded with VHS, hosted as assets on the <code>media-pr246</code> release tag — nothing binary is committed to the repo.</sub>

PR #249, opened 2026-08-28
feat(mermaid): render mermaid diagrams in the TUI

dimetron:
Adds internal/mermaid, a terminal renderer for 17 Mermaid diagram types,
adapted from github.com/aaronsb/mmaid-go (MIT) at 2d6873d. Full provenance,
the reasons it is copied rather than imported, and every local change are in
internal/mermaid/ATTRIBUTION.md.

Copied rather than imported because the upstream engine lives entirely under
its own internal/: a dependency can call exactly one function, and the width
override, the styled-cell accessor and the fixes below are all unreachable
from outside the module.

Added:

- WithWidth(cols), which upstream exposed only through its CLI. A library
  caller was left with the terminal's own width — wrong for any pane narrower
  than the terminal.
- diagram.WithWidthScope, pinning the package-level width fourteen renderers
  read directly, under a mutex. Threading it through each renderer as a
  parameter remains undone.
- RenderCells, returning styled cells so a caller applies its own palette.
- tui.RenderMermaid, drawing those cells with lipgloss and the active Palette
  so a diagram matches the pane instead of carrying foreign ANSI. It returns
  "" when output would exceed the pane, since a terminal cannot scroll
  sideways and a truncated diagram is worse than none.
- cmd/pi-mermaid, for eyeballing output without starting the TUI. Its -check
  mode batch-renders a directory and reports failures.

Fixed, each with a regression test:

- Non-determinism in column-slack distribution: ColWidths is a map, so the
  leftover columns went wherever iteration order landed and the same diagram
  rendered differently on successive calls. In a TUI that re-renders per
  frame this reads as flicker.
- Non-determinism in subgraph edge anchoring: members equidistant from a
  subgraph's center tied, and map order broke the tie, so an arrow left A1 on
  one render and A2 on the next.
- Panic on a nested block: entries the column pass never sized were indexed
  past the end of colX.
- ASCII mode leaked Unicode: shape indicators and dashed lines were hardcoded
  to their Unicode glyphs, so --ascii emitted them regardless.

Testing: 437 deduplicated inputs in testdata/corpus — 358 mermaid-example
fences from the 17 upstream mermaid syntax docs, 80 fixtures from
fasouto/termaid — each with an approved golden at width 100. Only the inputs
were taken; upstream's expected outputs are the Python renderer's, so every
golden here was generated and reviewed locally. Regenerate with -update and
read the diff before committing: a golden is an approval, not a snapshot
taken on trust.

Alongside the goldens the corpus asserts no panics, determinism across
repeated renders, valid UTF-8 with no control characters, width monotonicity,
and that ASCII mode introduces no non-ASCII runes.

419 of 437 render a diagram, 18 record genuine parser gaps in ER, class and
gantt syntax. Statement coverage is 90.7%, no package below 85%.


dimetron:
Closing in favour of a fresh PR against the current `main` — same branch, same commits, with a description covering all four commits rather than just the first. No review history is lost here; the only comment was Codecov's. Successor linked below.

dimetron:
Superseded by #250.

PR #250, opened 2026-08-28
feat(mermaid): render mermaid diagrams in the TUI

dimetron:
> **⚠️ This PR now also closes a critical `pi serve` vulnerability.** It began as the mermaid renderer; the security work was added at the author's request rather than split out. If review bandwidth is limited, review the security commit first — it is independent of everything else here.

## Security: `pi serve` was an unauthenticated remote shell

`pi serve` bound every interface by default, handed its pairing token to any anonymous caller, and approved that token without authentication. Three HTTP requests from anywhere on the network yielded an interactive agent session — hence its `bash` tool, hence arbitrary code execution as the user — with every provider API key inherited via `os.Environ()`. Each leg was already demonstrated by the repository's own tests; only their composition was new.

Findings and evidence are written up in `specs/issues/security-review.md`, with a `file:line` citation for every claim.

| Link | Fix |
|---|---|
| Bound `:8765` (all interfaces) | `DefaultAddr = "127.0.0.1:8765"`; remote access is opt-in via `--addr 0.0.0.0:8765`, and the banner tells the operator when pairing another device needs it |
| `GET /api/pair` returned the token, unauthenticated | The wire type carries `{code, qr}` only — the field is **deleted**, not omitted, so a field added later cannot leak by accident. A new internal `activePair` keeps the token for the operator's own banner |
| The token also travelled **inside the QR image** | The QR is base64'd into that same unauthenticated response, so stripping the JSON field alone would have left it readable. The QR payload no longer carries a token |
| `Approve` had no attempt limiting | Five wrong codes destroy every pending pair and lock submissions for a minute. Minting a new pair deliberately does **not** reset the budget — otherwise an attacker calls `/api/pair` for a fresh one; there is a test for that bypass |
| `CheckOrigin: return true` | Origin must match the request's own host. A **missing** Origin is accepted, since non-browser clients legitimately send none and a web page cannot suppress it; `null` from a sandboxed iframe is rejected |
| `pi_token` cookie had no `SameSite` | `SameSite=Strict`, `HttpOnly` retained |

The legacy `Server` in `handlers.go` carried duplicates of both the token leak and the open `CheckOrigin`, outside the review's citations; both are fixed.

Tests that asserted the old insecure behavior are **updated rather than deleted**, so a regression stays visible — the pair-response test decodes into a map, because a typed decode would hide a leaked field, and asserts the token appears neither in the raw body nor in the decoded QR image.

**Two items are flagged rather than fixed**, and need a product decision:
- The `?token=` query fallback still authenticates, and `pair.html`'s deep link depends on it.
- Attempt limiting is manager-global, not per-source: a wrong guess matches no pair, and a per-source counter behind a proxy falls to a forged `X-Forwarded-For`. The tradeoff is that an attacker can nuisance-lock pairing for a minute at a time on a now loopback-bound service.

**One externally visible behavior change:** if the mobile app scans a *token* out of the QR, that flow must now submit the code instead. `ParseQRData` still accepts the legacy `code:token` form so an older client degrades rather than breaks, but this could not be verified against a real client.

---

Renders Mermaid diagrams as terminal art in the TUI chat pane. Closed ` ```mermaid ` fences in an assistant reply are drawn as diagrams instead of printed as source; anything the renderer cannot draw, or that will not fit the pane, falls back to showing the fence unchanged.

Supersedes #249 (same branch, no review history lost — only a Codecov bot comment).

## What is here

**`internal/mermaid`** — a terminal renderer for 17 diagram types, adapted from [aaronsb/mmaid-go](https://github.com/aaronsb/mmaid-go) (MIT) at `2d6873d`, which is itself a Go reimplementation of [fasouto/termaid](https://github.com/fasouto/termaid) (MIT). Full provenance, the reasons it is copied rather than imported, and every local change are in `internal/mermaid/ATTRIBUTION.md`.

It is copied rather than imported because the upstream engine lives entirely under its own `internal/`: a dependency can call exactly one function, and the width override, the styled-cell accessor and the fixes below are all unreachable from outside that module. Upstream has zero third-party dependencies, and this copy keeps that — it adds nothing to `go.mod`.

**`internal/tui`** — `RenderMermaid` draws the renderer's styled cells with lipgloss and the active `Palette`, so a diagram matches the pane instead of carrying foreign ANSI. `RenderMarkdown` splits a reply into markdown and fence segments; prose still goes through glamour and art is spliced in afterwards, because glamour renders escape bytes as literal text.

**`cmd/pi-mermaid`** — a small CLI for eyeballing output without starting the TUI. Its `-check` mode batch-renders a directory and reports failures, which is how the corpus gets triaged after an engine change.

## Additions to the upstream engine

- `WithWidth(cols)`, which upstream exposed only through its CLI. A library caller was otherwise left with the terminal's own width — wrong for any pane narrower than the terminal.
- `diagram.WithWidthScope`, pinning under a mutex the package-level width that fourteen renderers read directly. Threading it through each renderer as a parameter is the real fix and remains undone.
- `RenderCells`, returning styled cells so a caller applies its own palette.

## Bugs fixed in the adapted code

Each has a regression test.

- **Two sources of non-determinism.** Column-slack distribution iterated a map, so leftover columns landed wherever iteration order put them and the same diagram rendered differently call to call. Subgraph edge anchoring broke equidistant ties by map order, so an arrow left `A1` on one render and `A2` on the next. In a pane that re-renders per frame both read as flicker.
- **Panic on a nested `block`** — grid entries the column pass never sized were indexed past the end of `colX`.
- **ASCII mode leaked Unicode** — shape indicators and dashed lines were hardcoded to their Unicode glyphs, so `--ascii` emitted them regardless.
- **`<br/>` corrupted nodes.** The shape patterns match raw punctuation and the tag carries a `/` and a `>`, so in `n["a<br/>b"]` the asymmetric `>`…`]` pattern matched inside the tag: the ID parsed as `n["a<br/` and the box came out a parallelogram. Folding now happens before shape matching.
- **Subgraph titles rendered as raw source.** Mermaid accepts both `subgraph SUB [Title]` and `subgraph SUB[Title]`, and models overwhelmingly emit the second; the pattern required the space, so a group was captioned `SUB["Title"]`, brackets and quotes included.
- **The fallback fence shimmered.** Chroma ships no mermaid lexer, so glamour guessed the language from content — a guess decided by map order. On a real 171-line diagram a ` ```mermaid ` fence differed between renders 48 times out of 60. The pane repaints on every blink tick, so the block quietly changed colour. The fallback now names a lexer Chroma has.

## Styling

Boxes, connectors and groups previously took three near-identical greys. Each role now reads by function: a box is chrome and stays back in neutral grey; its text takes full-strength foreground; a connector takes a hue rather than a grey, which is what lets the eye trace an edge; a group's border and title share the accent, tying a caption to the box it names; an edge label takes a warm tone distinct from both.

## Width

The engine treats width as a fill target it may exceed, so output is measured and rejected when it overflows — a terminal cannot scroll sideways, and a truncated diagram is worse than none. A diagram too wide as written is retried in the perpendicular orientation first, which is the largest single lever: a real 25-node graph measures 547 columns as `graph TD` and 131 as `LR`.

Known limit: a genuinely large diagram still falls back. A 44-node, 9-subgraph example needs 207 columns minimum even after the retry, and that is content-driven rather than a setting. Making such diagrams fit would need a compressing layout pass in `layout/grid.go`.

## Testing

`testdata/corpus/` holds 439 deduplicated inputs — 358 `mermaid-example` fences from the 17 upstream mermaid syntax docs, 80 fixtures from `fasouto/termaid`, and hand-added regressions — each with an approved golden at width 100. Only upstream's *inputs* were taken; its expected outputs are the Python renderer's, so every golden here was generated and reviewed locally.

Regenerate with `go test ./internal/mermaid -update` and read the diff before committing: a golden is an approval, not a snapshot taken on trust.

Alongside the goldens the corpus asserts no panics, determinism across repeated renders, valid UTF-8 with no control characters, width monotonicity, and that ASCII mode introduces no non-ASCII runes.

419 of 437 upstream cases render a diagram; 18 record genuine parser gaps in ER, class and gantt syntax. Statement coverage is 90.7% with no package below 85%. `go vet` and `golangci-lint` are clean.

Worth noting for review: both the `<br/>` and the subgraph-title bugs were invisible to the corpus — it contained no subgraph bracket labels and no `<br/>` in a flowchart label. Both are now corpus cases, so the gap that hid them is closed rather than worked around.


dimetron:
## govulncheck — run locally on this branch

```
Go: go1.27.0   Scanner: govulncheck@v1.7.0
DB: https://vuln.go.dev   updated 2026-08-28 14:47:45 UTC
```

**Result: no finding has a fix available. Gate passes (exit 0).**

| ID | Module | Fixed in |
|---|---|---|
| GO-2025-3557 | `github.com/ollama/ollama` | N/A |
| GO-2025-3558 | `github.com/ollama/ollama` | N/A |
| GO-2025-3559 | `github.com/ollama/ollama` | N/A |
| GO-2025-3582 | `github.com/ollama/ollama` | N/A |
| GO-2025-3689 | `github.com/ollama/ollama` | N/A |
| GO-2025-3695 | `github.com/ollama/ollama` | N/A |
| GO-2025-3824 | `github.com/ollama/ollama` | N/A |
| GO-2025-4251 | `github.com/ollama/ollama` | N/A |
| GO-2026-5750 | `github.com/ollama/ollama` | N/A |
| GO-2026-5932 | `golang.org/x/crypto` | N/A |

Ten findings, none actionable: there is no released version to upgrade to for any of them.

### Why these are reported but not failed on

The nine `ollama` entries all describe the **server** — authentication on model-management endpoints, GGUF parsing, inference denial-of-service. pi-go links only the HTTP client: `go list -deps` shows `api`, `auth`, `envconfig`, `format`, `internal/orderedmap`, `types/model` and `version`, and no `server/`, `llm/` or `fs/`.

govulncheck still prints call traces for them because the advisories carry no `ecosystem_specific` symbol scoping, so it matches the whole module and treats every exported symbol as vulnerable — which is why an identical trace block repeats for all eight symbol-level entries. The traces are real calls into `api.Client.Chat` and friends; the vulnerabilities they are attributed to are not in that code.

`GO-2026-5932` (`golang.org/x/crypto`) is module-level only — govulncheck found no call path to the vulnerable symbol.

### What this PR changes about it

Before: nothing ran a vulnerability scan. `make check-cve` existed but no workflow invoked it, and it opens with `go mod tidy -v`, which rewrites tracked files.

Now there is a `Vulncheck` job, gated so it **fails only on findings that name a fixed version**. Failing on unfixable ones would leave the check red on every commit until an upstream maintainer cuts a release, and a permanently red check is one nobody reads — the actionable findings would drown with it. Unfixed findings stay visible, so a fix appearing upstream turns one into a build failure at exactly the moment someone should act.

The gate is Go (`hack/vulngate`), reproducible with `make vulncheck`, and tested — including that a fixable finding still fails when unfixable ones are present, and that a finding whose advisory body is missing fails rather than passes.

### Worth a follow-up, not in this PR

Dropping the `ollama` dependency removes all nine of its advisories permanently and takes 41 modules out of the graph. pi-go calls five endpoints — `Chat`, `Embed`, `List`, `Show`, `NewClient` — which is a couple of hundred lines of local HTTP client.

PR #268, opened 2026-08-30
feat(session): rsync-safe session storage and lossless remote merge

dimetron:
Make the sessions tree safe to sync with rsync and add a merge path that
pulls remote sessions without ever losing local data.

- writeMeta and saveBranches now write via temp+rename (writeFileAtomic),
  so rsync never copies a torn meta.json or branches.json. events.jsonl
  and trajectory.atif.json were already atomic.
- internal/session/merge.go: MergeRemoteSessions folds a staged remote
  tree into the local one. events.jsonl is unioned by event ID (the longer
  file wins when one side is a strict continuation, otherwise dedupe by ID
  and order by timestamp, preserving raw lines); meta.json keeps the newer
  updatedAt; derived files follow the events winner; branches/ merge per
  branch. Local-only sessions are never touched and nothing is deleted.
  Non-session dirs (archive/, backup/, .idea/) are skipped.
- hack/session-merge: thin CLI over the merge with --dry-run.
- scripts/sync-sessions.sh: rsync remote sessions into a staging dir
  (raising the remote fd limit, excluding junk/temp files) then merge.

### DIFF
--- a/internal/session/merge.go
+++ b/internal/session/merge.go
@@ -10,9 +10,7 @@ import (
 	"io/fs"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
-	"time"
 )
 
 // MergeOptions controls MergeRemoteSessions.
@@ -35,17 +33,22 @@ type MergeReport struct {
 // place. Sessions that exist only on the remote side are copied in. Sessions
 // present on both sides are merged:
 //
-//   - events.jsonl: union by event ID. If one side is a prefix of the other
-//     (the common "one machine continued the session" case) the longer file
-//     wins; otherwise events are deduplicated by ID and ordered by timestamp.
+//   - events.jsonl: the local file is never rewritten — remote-only events are
+//     appended. If one side is a strict continuation of the other (the common
+//     "one machine continued the session" case) the missing tail is appended;
+//     otherwise the remote events whose IDs are absent locally are appended.
+//     Appending (rather than replace) means a pi still writing to the session
+//     cannot lose its events to the merge: both processes only ever add lines,
+//     at worst interleaved.
 //   - meta.json: the copy with the newer updatedAt wins.
 //   - trajectory.atif.json / acp.jsonl: the winning side's copy wins.
-//   - branches/: per-branch events are merged with the same union rule;
-//     branches.json metadata follows the events winner.
+//   - branches/: per-branch events are merged with the same append rule;
+//     branches.json branch entries are merged by name, preserving local-only
+//     branches.
 //
 // Local-only sessions are never touched, and nothing is ever deleted. All
-// writes are atomic (temp + rename), so an interrupted merge leaves each file
-// either old or new, never torn.
+// writes are atomic (temp + rename) or pure appends, so an interrupted merge
+// leaves each file either old or new, never torn.
 func MergeRemoteSessions(localDir, remoteDir string, opts MergeOptions) (MergeReport, error) {
 	var report MergeReport
 	entries, err := os.ReadDir(remoteDir)
@@ -66,7 +69,12 @@ func MergeRemoteSessions(localDir, remoteDir string, opts MergeOptions) (MergeRe
 			continue
 		}
 		localSession := filepath.Join(localDir, id)
-		if _, err := os.Stat(filepath.Join(localSession, "meta.json")); os.IsNotExist(err) {
+		exists, err := localSessionExists(localSession)
+		if err != nil {
+			report.Errors = append(report.Errors, fmt.Sprintf("%s: %v", id, err))
+			continue
+		}
+		if !exists {
 			if err := r.copyDir(localSession, remoteSession); err != nil {
 				report.Errors = append(report.Errors, fmt.Sprintf("%s: %v", id, err))
 				continue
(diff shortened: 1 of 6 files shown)


dimetron (commented):
I found several loss-safety gaps in the new merge path. The most important are local event overwrite, concurrent append loss, and branch metadata replacement.

— Codex review

internal/session/merge.go
```diff
@@ -0,0 +1,390 @@
...
+		id := entry.Name()
+		remoteSession := filepath.Join(remoteDir, id)
+		// Only real session dirs are synced. The sessions root also holds
+		// non-session dirs (archive/, backup/, .idea/) that must not be
+		// treated as sessions.
+		if !isSessionDir(remoteSession) {
+			continue
+		}
+		localSession := filepath.Join(localDir, id)
+		if _, err := os.Stat(filepath.Join(localSession, "meta.json")); os.IsNotExist(err) {
```
dimetron: [P1] Treat an events-only local session as existing

This checks only for `meta.json`, although `isSessionDir` below defines a session as having `meta.json` or `events.jsonl`. If a legacy or interrupted local session has `events.jsonl` but no metadata, `copyDir` takes the new-session path and overwrites its local events with the remote file instead of unioning them. That violates the lossless-local-data guarantee. Detect either session marker (and handle non-ENOENT stat errors) before choosing copy versus merge.

internal/session/merge.go
```diff
@@ -0,0 +1,390 @@
...
+		if err := r.writeFileAtomic(filepath.Join(localDir, "events.jsonl"), rData, 0o644); err != nil {
+			return "", err
+		}
+		return "remote", nil
+	case rMissing:
+		return "local", nil
+	}
+	merged, winner := mergeEventsData(lData, rData)
+	if winner != "local" {
+		if err := r.writeFileAtomic(filepath.Join(localDir, "events.jsonl"), merged, 0o644); err != nil {
```
dimetron: [P1] Prevent appends from being lost between read and rename

The local events are read earlier and then replaced here without coordinating with `FileService.AppendEvent`, which appends to the same path. An event appended after the read (or through an fd opened before this rename) is discarded when this snapshot replaces the file. Atomic rename avoids torn output, but it does not make this read-modify-write lossless. Please add cross-process locking or a compare/retry protocol that covers both appends and merge replacement.

internal/session/merge.go
```diff
@@ -0,0 +1,390 @@
...
+
+// mergeEventsData merges two events.jsonl payloads. It returns the merged
+// bytes and the winner: "local" (keep local), "remote" (remote is a strict
+// continuation), or "merged" (both sides contributed unique events).
+func mergeEventsData(local, remote []byte) ([]byte, string) {
+	l, err := parseRawEvents(local)
+	if err != nil {
+		return local, "local"
+	}
+	r, err := parseRawEvents(remote)
```
dimetron: [P2] Report an unparseable remote event file

A parse failure here silently returns the local payload as a successful merge; the caller then increments `Merged` and reports no error even though all remote events were skipped. This also conflicts with the repository rule to wrap errors with context. Return the parse error with the side/path identified so the per-session error report tells the user that the sync was incomplete.

internal/session/merge.go
```diff
@@ -0,0 +1,390 @@
...
+	if err != nil {
+		return local, "local"
+	}
+	if len(l) == 0 {
+		return remote, "remote"
+	}
+	if len(r) == 0 {
+		return local, "local"
+	}
+	if isPrefix(l, r) {
```
dimetron: [P2] Require a genuinely longer strict continuation

`isPrefix(l, r)` is also true when both ID sequences have equal length. Equal histories are therefore classified as a remote continuation, causing remote raw lines, derived files, and branch metadata to replace local versions even though the PR promises that only the longer strict continuation wins. Require `len(l) < len(r)` here (and the symmetric inequality below); equal sequences should follow the non-continuation policy.

internal/session/merge.go
```diff
@@ -0,0 +1,390 @@
...
+			if err := r.copyDir(lBranch, rBranch); err != nil {
+				return err
+			}
+			continue
+		}
+		if _, err := r.mergeEventsFiles(lBranch, rBranch); err != nil {
+			return err
+		}
+	}
+	if winner == "remote" {
```
dimetron: [P1] Merge branch metadata instead of replacing it from the root winner

The per-branch winners above are discarded, while this decision uses the unrelated root-events winner. If a remote-only branch was just copied and the root stays local, it is absent from local `branches.json` and cannot be selected. Conversely, a remote root continuation replaces the whole map and drops local-only branch entries. Merge the branch map by branch, preserving local-only entries and reconciling each shared branch using that branch's own event result.

internal/session/merge.go
dimetron: Fixed in 0d10697. The new-session detection now uses `localSessionExists` (checks both `meta.json` and `events.jsonl`), so an events-only local session is merged rather than overwritten. Resolving this thread.

internal/session/merge.go
dimetron: Fixed in 0d10697. events.jsonl is now append-only: remote-only events are appended with O_APPEND instead of the local file being rewritten, so a pi still writing to the session cannot lose events to the merge. Resolving this thread.

internal/session/merge.go
dimetron: Fixed in 0d10697. An unparseable remote events file now returns an error (`parsing remote events`) that surfaces as a per-session entry in the report naming the session, instead of silently skipping the remote events. Resolving this thread.

internal/session/merge.go
dimetron: Fixed in 0d10697. Continuation detection now requires a genuinely longer sequence (`len(l) < len(remote)` / `len(remote) < len(l)` before the prefix check), so equal-length histories are no longer classified as a continuation and do not trigger any write. Resolving this thread.

internal/session/merge.go
dimetron: Fixed in 0d10697. branches.json is now merged by branch name, preserving local-only branches and taking the larger per-branch head, instead of being wholesale replaced by the root events winner. Resolving this thread.

dimetron:
## govulncheck — branch `feat/session-sync-safe` @ 0d10697

**Result: PASS** — no findings with an available fix (the only gate that fails).

- Scanner: govulncheck (golang/vuln v1.7.0)
- Go: go1.27.0 darwin/arm64
- Findings: 10 reported, **ALL with no released fix** — ollama (9), golang.org/x/crypto (1). Nothing actionable, so the gate passes them through per repo policy.

Run: `make vulncheck` → `govulncheck -format json ./... | go run ./hack/vulngate`. DB checked 2026-08-30.

PR #274, opened 2026-08-31
feat/a2a streaming

dimetron:
- **feat(kagent): add image-digests.sh to identify per-arch release images**
- **feat(a2a): stream thinking and tool calls to the A2A client**

### DIFF
--- a/internal/a2a/server/stream.go
+++ b/internal/a2a/server/stream.go
@@ -1,0 +1,168 @@
+package server
+
+import (
+	"context"
+	"sync"
+
+	"github.com/a2aproject/a2a-go/v2/a2a"
+	"github.com/a2aproject/a2a-go/v2/a2asrv"
+	acp "github.com/coder/acp-go-sdk"
+)
+
+// ADK metadata keys, matching what kagent's ADK executor writes on A2A parts
+// (google.golang.org/adk/v2/server/adka2a/v2). The kagent UI classifies data
+// parts by shape ({name,args} vs {name,response}) rather than by these keys,
+// but keeping them lets ADK-based clients round-trip the parts.
+const (
+	adkMetaTypeKey    = "adk_type"
+	adkMetaThoughtKey = "adk_thought"
+)
+
+// a2aUpdater implements acpserver.SessionUpdater by translating ACP session
+// updates into A2A TaskArtifactUpdateEvents, mirroring how kagent's ADK
+// executor converts ADK session events to A2A artifacts:
+//
+//   - agent_message_chunk  → text artifact (append to one artifact per turn)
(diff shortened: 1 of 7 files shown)


internal/a2a/server/stream.go
```diff
@@ -0,0 +1,168 @@
...
+			ev.Artifact.Parts[0].SetMeta(adkMetaThoughtKey, true)
+		}
+		id = ev.Artifact.ID
+		if thought {
+			u.thoughtArtifactID = id
+		} else {
+			u.textArtifactID = id
+		}
+	} else {
+		ev = a2a.NewArtifactUpdateEvent(u.execCtx, id, a2a.NewTextPart(text))
```
dimetron, line 107: Streamed text and thought artifacts are never finalized. Both `NewArtifactEvent` and `NewArtifactUpdateEvent` leave `LastChunk` false, and the executor later emits only a completed status—nothing closes either saved artifact ID. ADK's A2A converter consequently treats these updates as partial (`Append || !LastChunk`) even after task completion, so kagent/ADK consumers can be left without a finalized assistant event. Please emit a terminal update for every created text/thought artifact before `Completed` (following the ADK executor's final-update pattern), or otherwise mark the actual final chunk with `LastChunk=true`.

internal/a2a/server/stream.go
```diff
@@ -0,0 +1,168 @@
...
+	})
+	part.SetMeta(adkMetaTypeKey, "function_call")
+	return u.emit(ctx, a2a.NewArtifactEvent(u.execCtx, part))
+}
+
+// emitToolEnd streams a tool result as a function_response data part.
+func (u *a2aUpdater) emitToolEnd(ctx context.Context, tu *acp.SessionToolCallUpdate) error {
+	u.mu.Lock()
+	name := u.toolNames[string(tu.ToolCallId)]
+	delete(u.toolNames, string(tu.ToolCallId))
```
dimetron, line 145: A `SessionToolCallUpdate` is not necessarily a terminal result. In this repo, `appendToParentLocked` sends content-only updates for every nested/subagent progress line. This code turns the first such update into a `function_response` with a nil response and deletes the saved name; subsequent progress and the real terminal result then fall back to `"tool"`. Please emit/delete the function response only when `tu.Status` is terminal (`completed`/`failed`) and handle or ignore nonterminal content updates separately.

dimetron (commented):
Reviewed `git diff main...HEAD` against the repository standards and the PR's A2A streaming contract. The implementation builds and its targeted race tests pass, but two protocol-level issues remain in streamed artifact finalization and tool-progress translation.

— Codex review

PR #279, opened 2026-08-31
feat(a2a): route kagent controller URLs through gRPC-Web

dimetron:
The kagent controller (v0.10.0-rc1) serves A2A over gRPC-Web on its HTTP
port (:8083), routed to AgentInstances via x-kagent-agent-instance-*
headers. The HTTP /api/a2a/kagent/<agent> path returns empty 200s, so
pi-go's JSON-RPC client got EOF against the URLs in ~/.pi-go/config.json.

Add a gRPC-Web transport for the a2a-go SDK, vendor the kagent
AgentInstanceService protos, and route kagent-style config URLs through
AgentInstance lifecycle (reuse READY or create) before sending A2A
messages. Streaming now skips non-terminal Tasks and lastChunk artifact
snapshots, and task extraction prefers artifacts over history so the
agent's reply is returned instead of the user's own message.


internal/tools/a2a_kagent.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	return resp.AgentInstances, nil
+}
+
+// createAgentInstance creates a new AgentInstance for the agent.
+func (c *kagentAPIClient) createAgentInstance(ctx context.Context, namespace, agent string) (*kagentapi.AgentInstance, error) {
+	req := &kagentapi.CreateAgentInstanceRequest{
+		Namespace:     namespace,
+		Harness:       agent,
+		AgentTemplate: agent,
```
dimetron, line 97: [P1] Populate the required `RequestId` field. The AgentInstance service validates `request_id` as a non-empty 1–128 character idempotency key, so this request is rejected as `InvalidArgument` whenever the client needs to create an instance. Generate a stable/unique value for this creation attempt (the deployment docs also show `request_id` as part of every create request).

dimetron (commented):
I found four actionable correctness issues in the new kagent lifecycle and streaming paths. The existing internal/tools tests pass, but they do not exercise these new paths.

— Codex review

internal/tools/a2a_kagent.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+	var resp kagentapi.CreateAgentInstanceResponse
+	if err := c.http.call(ctx, kagentAPIMethod("CreateAgentInstance"), req, &resp); err != nil {
```
dimetron: [P1] Use `callPath` for this non-A2A RPC. `call` always applies `methodPath`, so this currently posts to `/lf.a2a.v1.A2AService//kagent.api.v1alpha1.AgentInstanceService/CreateAgentInstance` instead of the vendored full method path. Listing works because it uses `callPath`; creation therefore always fails when no READY instance can be reused.

internal/tools/a2a.go
```diff
@@ -202,16 +212,28 @@ func appendStreamEvent(sb *strings.Builder, event a2a.Event) bool {
...
 	case *a2a.TaskStatusUpdateEvent:
 		// State change updates - check for terminal state
 		return e.Status.State.Terminal()
 
 	case *a2a.TaskArtifactUpdateEvent:
-		// Extract text from artifact parts
+		// Extract text from artifact parts. A lastChunk update carries the
+		// full accumulated artifact text, not a delta, so appending it would
+		// duplicate what the incremental updates already wrote.
+		if e.LastChunk {
```
dimetron: [P1] Do not discard every `LastChunk` event. In the A2A model, `LastChunk` only marks the final artifact chunk; `Append` determines whether its content is a delta or replacement. The SDK's own single-chunk and chunked servers put real output in an event with `LastChunk=true`, so this shared handler now returns an empty/truncated result for ordinary non-kagent streaming clients. Skip only the kagent replacement snapshot shape (for example `LastChunk && !Append` after prior chunks), while retaining final delta content.

internal/tools/a2a_grpcweb.go
```diff
@@ -0,0 +1,309 @@
...
+			yield(nil, fmt.Errorf("SendStreamingMessage: HTTP %d", res.StatusCode))
+			return
+		}
+
+		// gRPC-Web streams each response as its own data frame; the final
+		// frame is a trailer with grpc-status.
+		reader := newFrameReader(res.Body)
+		for {
+			flag, payload, err := reader.Next()
+			if errors.Is(err, io.EOF) {
```
dimetron, line 244: [P1] Treat EOF before the gRPC-Web trailer as an error. The trailer carries the authoritative `grpc-status`; accepting a bare EOF lets a dropped/truncated response look like a successful stream and can return partial output as completed. Track whether a trailer was received and only finish successfully after `grpc-status: 0` (the unary parser has the same missing-trailer acceptance after a data frame).

internal/tools/a2a_kagent.go
dimetron: Confirmed and fixed in 17be121.

The descriptor confirms it: `request_id` carries `min_len:1, max_len:128`, while `namespace`/`harness`/`agent_template` carry only `min_len:1`. An empty value is rejected with InvalidArgument.

`createAgentInstance` now sends `RequestId: newKagentRequestID()` — a fresh UUID per attempt, not a stable value, since reusing the key would make a genuine second create collapse into the first attempt's result. Covered by `TestCreateAgentInstancePathAndRequestID` (asserts the 1-128 range) and `TestNewKagentRequestIDIsUnique`.

internal/tools/a2a_kagent.go
dimetron: Confirmed and fixed in 17be121.

`call` unconditionally applies `methodPath`, which prepends `/lf.a2a.v1.A2AService/`, so the request went to `/lf.a2a.v1.A2AService//kagent.api.v1alpha1.AgentInstanceService/CreateAgentInstance`. Exactly as you noted, listing worked only because it already used `callPath`, so creation failed whenever no READY instance could be reused.

`createAgentInstance` now uses `callPath`. `TestCreateAgentInstancePathAndRequestID` asserts the request path is the bare AgentInstanceService method and that it no longer contains `lf.a2a.v1.A2AService`.

internal/tools/a2a.go
dimetron: Confirmed and fixed in 17be121, taking the rule you described.

The blanket `if e.LastChunk { return false }` returned an empty result for any server that sends its whole artifact in one frame — and that frame is legitimately marked LastChunk.

The handler now skips only the replacement snapshot: `LastChunk && !Append && already-wrote-this-artifact`. That is the shape kagent's ADK runtime and pi's own A2A server emit (deltas, then a full-text replacement); a final *delta* keeps its content, and a first-frame LastChunk keeps its content. Correlating across events needs state, so it moved into a `streamTextCollector` keyed by artifact ID.

Four cases cover it: `TestStreamCollectorKeepsSingleShotLastChunk`, `...SkipsReplacementSnapshot`, `...KeepsFinalDelta`, `...TracksArtifactsSeparately`.

internal/tools/a2a_grpcweb.go
dimetron: Confirmed and fixed in 17be121, including the unary parser you flagged.

Streaming took a bare `io.EOF` as a clean end, and `callPath` checked only `sawData` before unmarshalling, so both treated a dropped or truncated response as success.

Both now track whether the trailer frame arrived and fail with a shared `errMissingGRPCWebTrailer` when it did not; the non-zero `grpc-status` path is unchanged. `TestCallPathRejectsMissingTrailer` drives a server that writes a data frame and stops, and `TestCallPathAcceptsTrailer` is the positive case so the check cannot pass by rejecting everything.

dimetron:
## govulncheck — clean gate

Run on `feat/a2a-kagent-grpcweb-client` at 17be121, after the review findings were resolved.

```
Go: go1.27.0   Scanner: govulncheck@v1.7.0
DB: https://vuln.go.dev   DB updated: 2026-08-28 14:47:45 +0000 UTC
```

`make vulncheck` exits 0: **no findings with an available fix.** 10 findings are reported without failing, because none names a fixed version — 9 in `github.com/ollama/ollama` (GO-2025-3557, -3558, -3559, -3582, -3689, -3695, -3824, GO-2025-4251, GO-2026-5750) and GO-2026-5932 in `golang.org/x/crypto`. All pre-date this branch; it adds no new dependency with an actionable advisory.

This answer is only true for the database as of the date above.

dimetron:
## govulncheck — ✅ pass

Ran `make vulncheck` (govulncheck `-format json ./...` piped through `hack/vulngate`) against this branch.

| | |
|---|---|
| Scanner | govulncheck@v1.7.0 (Go go1.27.0) |
| DB | https://vuln.go.dev, last modified 2026-08-28T14:47:45Z |
| Findings with an available fix | **0** (gate exit 0) |

The gate fails only on findings that name a fixed version; findings with no released fix are printed but do not fail. None were reported. The `tmp/adk` package-resolution errors in the log are pre-existing (a separate module not part of the build) and do not affect the result.

PR #282, opened 2026-08-31
feat: add Nix flake and NixOS module

dimetron:
Adds Nix packaging for pi-go: a flake exposing the `pi` binary, a dev shell, and a NixOS module.

## What's here

- **`flake.nix`** — `buildGoModule` package for `x86_64-linux` and `aarch64-linux`, building `cmd/pi` only. Version and `BuildTag` are stamped via `ldflags` from `self.shortRev`.
- **`apps.default`** — `nix run` support.
- **`devShells.default`** — `go_1_26`, `git`, `gnumake`.
- **`nixosModules.default`** — `programs.pi-go.enable` / `programs.pi-go.package`, installing into `environment.systemPackages`.
- **`flake.lock`** pinned to `nixos-unstable`; README gains install/usage notes.

## Known issue — the `postPatch` block needs a follow-up

The package patches `go.mod` because nixpkgs' pinned Go is older than the repo's `go 1.27.0` directive. Two substitutions use `--replace-fail`, which aborts the build when a pattern is absent, and one of them no longer matches:

\`\`\`nix
--replace-fail 'google.golang.org/grpc v1.83.1 // indirect' 'google.golang.org/grpc v1.83.1'
\`\`\`

The branch is based on 4e9f3e2, where grpc sat in the indirect block. On current \`main\` (8edb09b) it is already a **direct** dependency with no \`// indirect\` comment, so this substitution finds nothing and fails the build. It should simply be dropped — main now does what it was compensating for. The \`go 1.27.0\` → \`go 1.26.5\` substitution is still correct and still needed.

Note that \`vendorHash\` will need recomputing once \`go.mod\` changes, and this branch has not yet been updated against current main.

## Not verified here

The flake has not been built on this machine — no Nix on the macOS host, and the source worktree lives inside the NixOS OrbStack VM. \`nix build .#default\` on Linux is the check that matters before merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,39 @@ syntax in prompts: `;` or a newline rather than `&&`. Installing
 [Git for Windows](https://git-scm.com/download/win) puts a `bash` on `PATH` and
 restores the bash behaviour.
 
+### NixOS / Nix
+
+The repository includes a flake that builds `pi-go` reproducibly and exposes a
+NixOS module. To install it in a NixOS configuration, add the repository as an
+input:
+
+```nix
+# flake.nix
+{
+  inputs.pi-go.url = "github:dimetron/pi-go";
+
+  outputs = { self, nixpkgs, pi-go, ... }: {
+    nixosConfigurations.my-host = nixpkgs.lib.nixosSystem {
+      system = "x86_64-linux";
+      modules = [
+        ./configuration.nix
+        pi-go.nixosModules.default
+      ];
+    };
+  };
+}
+```
+
+Enable it in `configuration.nix`:
+
+```nix
+{ programs.pi-go.enable = true; }
+```
+
+Then rebuild with `sudo nixos-rebuild switch --flake .`. For a one-off use,
+run `nix run github:dimetron/pi-go` or install it into a profile with
+`nix profile install github:dimetron/pi-go`.
+
 ### go install
 
 ```bash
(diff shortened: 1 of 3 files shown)


dimetron:
Pushed d8570cd: dropped the stale grpc substitution.

**Why it was fatal rather than cosmetic.** `substituteInPlace --replace-fail` exists so a stale pattern fails loudly instead of silently no-opping. Checked against current `main` (995adb7):

| pattern | matches in main's go.mod |
|---|---|
| `go 1.27.0` | 1 ✓ |
| `google.golang.org/grpc v1.83.1 // indirect` | **0** ✗ |

main now requires grpc directly at line 46 with no `// indirect` comment, so the substitution the branch added at 4e9f3e2 has nothing to match and aborts `nix build` outright. Dropping it is the fix — main does what it was compensating for. The `go 1.27.0 → 1.26.5` patch still matches exactly once and is retained.

## Still needed before this can merge

Two things I could not do from here, both requiring a machine with Nix:

1. **`vendorHash` is almost certainly stale.** It was computed against 4e9f3e2's dependency set. main has moved since — #280 dropped easyjson via a `replace` directive, #279 added `internal/kagentapi`, and #283 is proposing adk v2.2.0 → v2.3.0. Recompute with `nix build .#default` and paste the hash it reports.
2. **Confirm `go_1_26` is still the right pin.** The dev shell and the `1.26.5` directive patch both assume nixpkgs-unstable has no Go 1.27 yet. Worth re-checking against the current lock before merge.

Neither is verifiable on this macOS host — there's no Nix here, and the source worktree lives inside the NixOS OrbStack VM.

PR #285, opened 2026-08-31
feat(provider): add agentgateway provider

dimetron:
Adds `agentgateway` as a first-class provider. agentgateway is an
OpenAI-compatible LLM gateway, so the client is the OpenAI one pointed at the
gateway's base URL.

## Usage

```bash
pi --model agentgateway/agentgateway "draft release notes"
pi model list agentgateway
```

| Env var | Purpose |
|---|---|
| `AGENTGATEWAY_BASE_URL` | Gateway endpoint (default `http://localhost:4000`) |
| `AGENTGATEWAY_API_KEY` | Optional — only for a gateway that fronts itself with a key |

## Notes

**No API key required.** The gateway holds the upstream credentials, so
`requireRuntimeAPIKey` exempts the provider and `pi model list` reports
`none (local gateway)` rather than `MISSING`.

**Prefix beats suffix.** The `agentgateway/` prefix is matched *before* the
`:cloud`/`-cloud` suffix check in both `Resolve` and `autoDetectProvider`.
Gateway model IDs commonly carry a `-cloud` tag (e.g.
`deepseek-v4-flash:0731-cloud`), which would otherwise route them to Ollama.
`switchedModelName` re-adds the prefix for the same reason, so an in-session
model switch cannot silently move the model to Ollama.

Also wired through model listing (`GET /v1/models`, shared bearer envelope),
`pi ping` (including the `/v1`-suffixed base URL case), config API keys and
base URLs, and the OTel `gen_ai.provider.name` attribute.

## Verification

Run against a live agentgateway instance:

- `pi model list agentgateway` enumerates the gateway's models
- `pi ping --model agentgateway/agentgateway "2+2"` → replies `4` on both the
  non-streaming and streaming paths
- A print-mode turn with tools completes a tool call and returns the result

`make test` (full unit suite), `make lint` (0 issues), `go vet`, and
`gofmt` all clean.


dimetron:
## govulncheck

```
Go: go1.27.0 Scanner: govulncheck@v1.7.0 DB: https://vuln.go.dev DB updated: 2026-08-28 14:47:45 +0000 UTC  
```

No findings with an available fix. 10 findings have no released fix and are reported without failing the gate:

| ID | Module |
|---|---|
| GO-2025-3557 | github.com/ollama/ollama |
| GO-2025-3558 | github.com/ollama/ollama |
| GO-2025-3559 | github.com/ollama/ollama |
| GO-2025-3582 | github.com/ollama/ollama |
| GO-2025-3689 | github.com/ollama/ollama |
| GO-2025-3695 | github.com/ollama/ollama |
| GO-2025-3824 | github.com/ollama/ollama |
| GO-2025-4251 | github.com/ollama/ollama |
| GO-2026-5750 | github.com/ollama/ollama |
| GO-2026-5932 | golang.org/x/crypto |

All pre-existing on `main`; this branch adds no dependencies. True for the vulnerability database as of the scan date above.

internal/cli/ping.go
```diff
@@ -101,6 +103,8 @@ func pingEndpoint(providerName string) string {
 		return "/v1beta/models"
 	case "xai":
 		return "/v1/models"
+	case "agentgateway":
```
dimetron, line 106: [P1] Forward the optional agentgateway key on the health probe. `resolvePingCredentials` loads `AGENTGATEWAY_API_KEY`, but `setPingAuthHeaders` only emits bearer auth for `openai` and `xai`. A gateway configured to require the documented optional key therefore returns 401 from this new `/v1/models` probe, and `verdictUnauthorized` aborts before the model ping even though the later OpenAI client would use the key. Please include `agentgateway` in the bearer-auth branch and cover the keyed probe in a test.

internal/extension/tracing_test.go
```diff
@@ -311,14 +311,15 @@ func TestToolSpansAreExportedPerCall(t *testing.T) {
...
-		"opencode":  "opencode",
-		"":          "",
+		"openai":       "openai",
+		"anthropic":    "anthropic",
+		"gemini":       "gcp.gemini",
+		"mistral":      "mistral_ai",
+		"xai":          "x_ai",
+		"azure":        "azure.ai.openai",
+		"opencode":     "opencode",
+		"agentgateway": "agentgateway",
```
dimetron, line 321: [P2] Align this assertion with the telemetry contract stated by the PR. The PR promises an OTel `gen_ai.system` attribute, but `genAIProviderAttr` returns `gen_ai.provider.name`, and this test checks only the value, so it cannot catch the key mismatch. Either emit and assert `gen_ai.system=agentgateway`, or update the PR contract if the newer `gen_ai.provider.name` key is intentional.

internal/provider/agentgateway.go
```diff
@@ -0,0 +1,24 @@
...
+// NewAgentGateway creates an agentgateway model.LLM. agentgateway is an
+// OpenAI-compatible local gateway, so the client is the OpenAI one pointed at
+// the gateway's base URL. No API key is required: NewOpenAI substitutes a
+// dummy key whenever a base URL is set, and a gateway that needs one can still
+// be given AGENTGATEWAY_API_KEY.
+func NewAgentGateway(ctx context.Context, modelName, apiKey, baseURL string, opts *LLMOptions) (model.LLM, error) {
+	if baseURL == "" {
+		baseURL = agentgatewayDefaultBaseURL
+	}
+	return NewOpenAI(ctx, modelName, apiKey, baseURL, opts)
```
dimetron: [P3] Preserve provider context when construction fails. `NewOpenAI` can return transport/TLS configuration errors, and directly returning it here loses the fact that the failure occurred while constructing agentgateway. `CONTRIBUTING.md` requires errors to be wrapped with context; please wrap this as, for example, `fmt.Errorf("creating agentgateway client: %w", err)`.

internal/provider/list_models.go
```diff
@@ -80,6 +82,8 @@ func ListModels(ctx context.Context, providerName string, opts ListModelsOptions
 		return listXAIModels(ctx, opts)
 	case "openrouter":
 		return listOpenRouterModels(ctx, opts)
+	case "agentgateway":
```
dimetron, line 85: [P3] Update the exported `ListModels` documentation when adding this branch. Its comment still says the supported providers end with `ollama` (and omits this new provider), so callers reading the API contract cannot discover agentgateway support. Please update the list or make the comment non-enumerative.

internal/cli/model.go
```diff
@@ -200,6 +201,9 @@ func modelListBaseURL(providerName string, baseURLs map[string]string) string {
 	if baseURL == "" && providerName == "ollama" {
 		baseURL = "http://localhost:11434"
 	}
+	if baseURL == "" && providerName == "agentgateway" {
```
dimetron, line 204: [P3] Update the comment above this helper: it currently says only Ollama has a default, while this new branch gives agentgateway one too. Keeping that contract accurate matters because the distinction explains why these providers work without configuration.

dimetron (commented):
Reviewed `main...HEAD` against the PR contract and the repository standards. I found one runtime defect in keyed-gateway ping handling, one telemetry contract mismatch, and three smaller standards/documentation issues. Focused package tests completed with 2,661 passes; the sole failure was an unrelated `no space left on device` error while a config test created a temp file.

— Codex review

internal/cli/ping.go
dimetron: Confirmed and fixed in caaa790. `fetchJSON` in `list_models.go` already sent the bearer for `agentgateway`, so listing and ping disagreed — a keyed gateway 401'd on the probe and aborted before the model ping it exists to run. `setPingAuthHeaders` now includes `agentgateway` in the bearer branch, with a table case in `ping_helpers_test.go`.

internal/extension/tracing_test.go
dimetron: Good catch, but the defect is in my PR description, not the code. `genAIProviderAttr` emits `gen_ai.provider.name` deliberately — that is the current semconv key, and the existing assertions at `tracing_test.go:119/165/219` all use it. The test is right to check the value, since the key is already covered by those. I have corrected the PR body to say `gen_ai.provider.name`. No code change.

internal/provider/agentgateway.go
dimetron: Fixed in caaa790 — wrapped as `fmt.Errorf("creating agentgateway client: %w", err)`. Worth it here because the underlying error names the OpenAI client, which is confusing for someone who asked for agentgateway. Noting that `opencode.go:74` returns `NewOpenAI` unwrapped in the same way; leaving that alone as out of scope for this PR.

internal/provider/list_models.go
dimetron: Fixed in caaa790. Rather than adding `agentgateway` to the list, the comment is now non-enumerative — it had already gone stale on `openrouter` before this PR, so an enumeration that has drifted once will drift again.

internal/cli/model.go
dimetron: Fixed in caaa790 — the comment now names both Ollama and agentgateway, and says why they are the two with defaults (both expected to be running locally on a well-known port with no configuration).

PR #316, opened 2026-09-04
fix(ratelimit): pace Gemini input tokens against a rolling window, not a bucket

### LINKED ISSUE #315: ratelimit: token bucket lets bursty Gemini traffic blow the rolling 2M/min window mid-stream
## Problem

`internal/ratelimit` paces Gemini traffic to stay under the published
`generate_content_paid_tier_input_token_count` quota (2,000,000 input
tokens/min, `quotaId=GenerateContentPaidTierInputTokensPerModelPerMinute`).
The default budget is `1_800_000`/min (`defaults.go:34`), a 10% haircut.

The limiter is a **token bucket** (`rate.NewLimiter(rate.Limit(1.8M/60), 1.8M)`)
with a **full-window burst** — the comment in `New` says "spend it all at once
if you like, then wait." That design does not model the server's **rolling**
60-second window, so bursty traffic over-spends and gets a mid-stream 429.

## Evidence

Recent sessions (2026-09-03/04) hit:

```
transient error after partial response (not retrying): Error 429 ...
Quota exceeded for metric: generativelanguage.googleapis.com/generate_content_paid_tier_input_token_count,
limit: 2000000, model: gemini-3.8-flash
```

Two distinct gaps:

1. **Bucket ≠ rolling window.** A full-window burst lets pi-go dump ~1.8M
   tokens instantly, then again 60s later. The server's rolling window still
   holds the first burst when the second lands → 3.6M in one window → 429.
   The 10% haircut only covers the refill, not a second full burst.

2. **Mid-stream 429 is unrecoverable.** The limiter only gates the request
   *send* (`transport.go:67`, estimating input tokens from `ContentLength`).
   Once the stream is open, the limiter is out of the picture. A 429 that
   arrives mid-stream is "after partial response", so `agent/retry.go:70-74`
   correctly refuses to re-run the turn — the turn is lost.

## Proposed direction

A stream-aware limiter that paces *during* the stream, not just at send:

- **Rolling-window accounting** instead of (or in addition to) the token
  bucket, so a burst at t=0 still counts against the window at t=60.
- **Mid-stream pacing with delay**: when a streamed request is admitted, keep
  a running token ledger for the open stream and insert delays between chunks
  (or before the next request) so cumulative spend stays under the window.
  This is what turns a mid-stream 429 into a controlled slow-down.
- Keep the existing send-gate for the common case; the stream limiter is the
  second net that catches the bursty/mid-stream case.

## Scope

- `internal/ratelimit/` — add a rolling-window / stream-pacing primitive.
- `internal/provider/` — wire it into the Gemini streaming path.
- Tests: reproduce the burst-then-429 case (two full bursts 60s apart) and a
  mid-stream over-spend.

dimetron:
## Problem

`internal/ratelimit` paces Gemini traffic to stay under the published
`generate_content_paid_tier_input_token_count` quota (2,000,000 input
tokens/min). The token budget was a `rate.Limiter` with a **full-window burst**
— the comment in `New` said "spend it all at once if you like, then wait."

That is the wrong shape for a **rolling** 60-second window. A burst at t=0 still
counts at t=59 and only falls out at t=60, so a second full burst a minute later
lands while the first is still counted — 3.6M tokens in one window — and Gemini
answers with a mid-stream 429:

```
transient error after partial response (not retrying): Error 429 ...
Quota exceeded for metric: generativelanguage.googleapis.com/generate_content_paid_tier_input_token_count,
limit: 2000000, model: gemini-3.8-flash
```

A mid-stream 429 is unrecoverable: `agent/retry.go` refuses to re-run a turn
once it has yielded anything. The 10% haircut (1.8M default) only covers the
refill, not a second full burst.

## Fix

Replace the token bucket with a **rolling-window ledger** in
`internal/ratelimit`:

- Each admitted request records its token charge and admission time.
- Charges older than the window are dropped (they've fallen out of the server's
  rolling window).
- A request is **delayed** until its charge fits the remaining window — a second
  full burst is held until the first falls out, turning the mid-stream 429 into
  a controlled delay at admission.

The RPM budget stays a token bucket (a full-window burst is fine for request
*count*), and server-ordered `Backoff` is unchanged.

## Tests

- `TestLimiterHoldsSecondFullBurst` — the exact failure: a second full burst a
  minute after the first is now held, not admitted.
- `TestLimiterRollingWindowExpiry` — a charge at t=0 still counts at t=59, falls
  out at t=60.

All existing ratelimit/config/provider tests pass; `go build ./...` and
`go vet` are clean.

Fixes #315

### DIFF
--- a/internal/ratelimit/ratelimit.go
+++ b/internal/ratelimit/ratelimit.go
@@ -84,6 +84,15 @@ type spendEvent struct {
 	tokens int
 }
 
+// waiter is one request queued for token capacity. Waiters are served in FIFO
+// order so a large request cannot be starved by a stream of small ones.
+type waiter struct {
+	charge int
+	// ready is closed when this waiter becomes the head and should re-check
+	// whether it can be admitted.
+	ready chan struct{}
+}
+
 // Limiter paces requests against a rolling input-token window and/or a
 // request-per-minute budget, and absorbs server-ordered cooldowns.
 //
@@ -100,6 +109,8 @@ type Limiter struct {
 	tokenBurst int
 	// spend holds the token charges still inside the window, oldest first.
 	spend []spendEvent
+	// waiters is the FIFO queue of requests waiting for token capacity.
+	waiters []*waiter
 
 	mu        sync.Mutex
 	coolUntil time.Time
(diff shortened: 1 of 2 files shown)


internal/ratelimit/ratelimit.go
```diff
@@ -120,24 +143,59 @@ func (l *Limiter) Wait(ctx context.Context, inputTokens int) error {
...
+		for keep < len(l.spend) && l.spend[keep].at.Before(cutoff) {
+			keep++
+		}
+		l.spend = l.spend[keep:]
+
+		sum := 0
+		for _, e := range l.spend {
+			sum += e.tokens
+		}
+		if sum+charge <= l.tokenLimit {
```
dimetron: [P1] This admission check has no reservation or waiter ordering: when a large charge is waiting, any smaller caller that acquires `mu` first can consume the newly freed capacity. With continuous small requests on this shared limiter, a full-window request can be leapfrogged at every expiry and wait forever, even with a live context. The previous `rate.Limiter` reserved future capacity for waiters. Please queue waiters (or otherwise reserve capacity in arrival order) so concurrent subagents cannot starve an older large request.

internal/ratelimit/ratelimit.go
```diff
@@ -120,24 +143,59 @@ func (l *Limiter) Wait(ctx context.Context, inputTokens int) error {
...
+			l.spend = append(l.spend, spendEvent{at: now, tokens: charge})
```
dimetron, line 228: [P2] Check `ctx.Err()` before recording an immediately admissible charge. With a token-only limiter and available capacity, `Wait` currently appends to `spend` and returns nil even when the context was already canceled; the prior `rate.WaitN` explicitly rejected that case. This can charge budget for a request that should never be sent and violates the documented “until ctx is canceled” contract. A regression test using a pre-canceled context would pin the behavior.

internal/ratelimit/ratelimit_test.go
```diff
@@ -84,6 +84,55 @@ func TestLimiterBlocksOnRequestBudget(t *testing.T) {
...
+// token bucket with a full-window burst would over-spend and 429. The rolling
+// window delays the second burst until the first falls out.
+func TestLimiterHoldsSecondFullBurst(t *testing.T) {
+	t.Parallel()
+	lim := New(Limits{InputTokensPerMinute: 1000})
+	if err := lim.Wait(context.Background(), 1000); err != nil {
+		t.Fatalf("first burst Wait: %v", err)
+	}
+
+	// A second full burst immediately must wait for the first to fall out of
```
dimetron, line 99: [P2] This test does not distinguish the new rolling window from the old token bucket: immediately after consuming a full burst, the old bucket also blocks another full burst for longer than 50 ms. To reproduce the bug, advance an injected clock to just before 60 seconds and request a partial charge that the bucket would have refilled but the rolling ledger must still reject. That would make the test fail on the implementation being replaced instead of only asserting generic exhaustion.

dimetron (commented):
Reviewed `main...HEAD` for rolling-window correctness, concurrency, cancellation, and regression coverage. The targeted package passes `go test -race ./internal/ratelimit` and `go vet ./internal/ratelimit`; the inline comments identify the remaining actionable issues.\n\n— Codex review

internal/ratelimit/ratelimit.go
dimetron: Fixed in 26cba3e. Resolving.

internal/ratelimit/ratelimit.go
dimetron: Fixed in 26cba3e. Resolving.

internal/ratelimit/ratelimit_test.go
dimetron: Fixed in 26cba3e. Resolving.

PR #324, opened 2026-09-06
fix(provider): send max_tokens instead of max_completion_tokens for agentgateway

dimetron:
## Summary

Ollama only understands the legacy `max_tokens` field; the newer `max_completion_tokens` (which the openai-go SDK sends by default) is silently ignored. When pi-go sends `max_completion_tokens=64000` through agentgateway to an Ollama-backed model, Ollama falls back to its own 65536-token default, truncating long turns at a different limit than the one pi-go asked for — and the cut can land mid-generation with no content at all.

**Session 260906-1320** was the reported case: `ollama-deepseek` via agentgateway, 116K prompt tokens, 65536 output tokens, empty `parts: []`, `finishReason: MAX_TOKENS`. The model hit Ollama's 65536 default because pi-go's 64000 cap was sent in a field Ollama doesn't read.

## Fix

Add `useLegacyMaxTokens` to `openaiModel`, set it for the agentgateway provider, and send `MaxTokens` (`max_tokens`) instead of `MaxCompletionTokens` (`max_completion_tokens`) on the Chat Completions wire when it is set.

No agentgateway config change needed — the gateway already forwards `max_tokens` correctly; the issue was entirely that pi-go was not sending the field Ollama reads.

## Files changed

- `internal/provider/provider.go` — add `UseLegacyMaxTokens` to `LLMOptions`
- `internal/provider/openai.go` — add `useLegacyMaxTokens` field, wire from `LLMOptions`
- `internal/provider/openai_completions.go` — send `params.MaxTokens` when set, `params.MaxCompletionTokens` otherwise
- `internal/provider/agentgateway.go` — force `UseLegacyMaxTokens = true`
- `internal/provider/functioncall_args_test.go` — new test `TestAgentGatewaySendsLegacyMaxTokens`

## Verification

- `go build ./...` ✅
- `go vet ./internal/provider/...` ✅
- `golangci-lint` ✅ (0 issues)
- `go test ./internal/provider/...` ✅ (all tests pass including new test)

### DIFF
--- a/internal/provider/agentgateway.go
+++ b/internal/provider/agentgateway.go
@@ -17,10 +17,20 @@ const agentgatewayDefaultBaseURL = "http://localhost:4000"
 // the gateway's base URL. No API key is required: NewOpenAI substitutes a
 // dummy key whenever a base URL is set, and a gateway that needs one can still
 // be given AGENTGATEWAY_API_KEY.
+//
+// UseLegacyMaxTokens is forced on because agentgateway typically routes to
+// Ollama, whose API only understands the legacy max_tokens field — the newer
+// max_completion_tokens is silently ignored, so the model runs unbounded and
+// hits Ollama's own 65536-token default. Sending max_tokens lets pi-go's cap
+// (defaultOaiMaxOutputTokens = 64000) reach the model instead.
 func NewAgentGateway(ctx context.Context, modelName, apiKey, baseURL string, opts *LLMOptions) (model.LLM, error) {
 	if baseURL == "" {
 		baseURL = agentgatewayDefaultBaseURL
 	}
+	if opts == nil {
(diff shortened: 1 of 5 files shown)


dimetron (commented):
Reviewed `git diff dc4fc43...48db789`.

Standards: no actionable findings.

Spec: no actionable findings against the PR description. The agentgateway path selects `max_tokens`, the normal OpenAI path retains `max_completion_tokens`, and output-limit precedence is preserved. Checked gateway token-limit normalization for OpenAI backends in the local upstream source, including tag v1.5.0.

Validation: `go test -mod=readonly ./internal/provider/...` passed (1,289 tests). No live upstream model requests were made.

— Codex review

PR #350, opened 2026-09-18
feat(vscode): improve Pi-Go chat diagnostics and ping

dimetron:
## What changed

Improves the Pi-Go VS Code extension chat experience and troubleshooting flow:

- Adds a monochrome Ping icon that runs `pi ping` without interrupting the active session.
- Shows compact provider/model health output with credential redaction, configured ping flags, and correct failure handling when the CLI exits zero after an unhealthy probe.
- Replaces opaque internal errors with a high-contrast diagnosis card containing raw details, workspace/launch context, recovery steps, and links to settings/output.
- Keeps thinking and tool output collapsed by default to reduce transcript overload.
- Adds Pi-Go walkthrough/getting-started UI, blue gopher + pi branding, and removes the session-tree reveal error.
- Skips native-session registration cleanly unless VS Code was launched with `--enable-proposed-api pi-go.pi-go-vscode`; the dedicated chat remains available in Stable.
- Adds repository guidance requiring isolated worktrees and instruction-file discovery before edits.

## Validation

- `cd vscode && bun run test` — 131 tests passed.
- TypeScript host and webview typechecks passed.
- VSIX packaged and reinstalled into VS Code Stable as `pi-go.pi-go-vscode@0.3.0`.
- Installed-profile smoke test passed: extension activation, ACP startup, and prompt roundtrip.
- `go test ./...` ran 11,233 tests successfully but reports four pre-existing build failures under `tmp/sdk/adk/adk-samples/...` because `google.golang.org/adk/agent` is unavailable in those sample modules.


vscode/src/ping.ts
```diff
@@ -0,0 +1,79 @@
...
+    .replace(/\x1b\[[0-?]*[ -/]*[@-~]/g, "")
+    .replace(/(API\s*Key\s*:\s*).*/gi, "$1[redacted]")
+    .replace(/(Authorization\s*:\s*).*/gi, "$1[redacted]")
+    .replace(/(Bearer\s+)[A-Za-z0-9._~+/=-]+/gi, "$1[redacted]")
+    .trim();
+}
+
+export function formatPiPingResult(result: PiGoPingProcessResult): PiGoPingResult {
+  const output = compactPingOutput(sanitizePiPingOutput([result.stdout, result.stderr].filter(Boolean).join("\n")));
+  if (result.exitCode === 0) {
```
dimetron: [P1] Treat unhealthy CLI output as a failed ping

`pi ping` writes diagnostics to stderr, and `runPing` returns nil after `reportHTTPVerdict` prints 401/404/5xx failures (for example, it prints `Skipped — endpoint not reachable` for a 500). This maps every exit code 0 to success, so the new button can show a successful ping while the provider is down. Parse an explicit success marker, or make the CLI return non-zero for an unhealthy verdict.

vscode/src/ping.ts
```diff
@@ -0,0 +1,79 @@
...
+  ok: boolean;
+  title: string;
+  detail: string;
+}
+
+/** Remove ANSI styling and values that should never be copied into chat. */
+export function sanitizePiPingOutput(output: string): string {
+  return output
+    .replace(/\x1b\[[0-?]*[ -/]*[@-~]/g, "")
+    .replace(/(API\s*Key\s*:\s*).*/gi, "$1[redacted]")
```
dimetron, line 25: [P1] Redact URL query credentials before rendering ping output

The Go ping implementation adds Gemini credentials as `?key=...`; transport errors can echo the full request URL, and the compaction filter keeps lines containing `HTTP FAILED`/`error`. This sanitizer only redacts labelled API-key/Authorization/Bearer values, so a failed URL-based request can put the raw key in the webview. Redact sensitive query parameters (and similar URL credentials) before forwarding output.

vscode/src/errorInfo.ts
```diff
@@ -0,0 +1,119 @@
...
+function codeOf(error: unknown): string | undefined {
+  if (typeof error !== "object" || error === null || !("code" in error)) return undefined;
+  return String((error as { code: unknown }).code);
+}
+
+export function explainPiGoError(error: unknown, config: PiGoLaunchConfig): PiGoErrorInfo {
+  const raw = messageOf(error);
+  const lower = raw.toLowerCase();
+  const code = codeOf(error);
+  const launch = `${config.command} ${config.args.join(" ")}`.trim();
```
dimetron: [P1] Do not render raw configured args in diagnosis details

`pi-go.args` is arbitrary and can contain `--header key=value` or provider credentials. This string is copied verbatim into a user-visible error card and native chat response, so a startup error can expose secrets. Keep useful launch context, but redact secret-bearing flags/values or omit the args entirely.

vscode/src/errorInfo.ts
```diff
@@ -0,0 +1,119 @@
...
+  const detail = `${raw}\nLaunch: ${launch}\nWorkspace: ${config.cwd}`;
+
+  if (code === "ENOENT" || lower.includes("enoent") || lower.includes("not found")) {
```
dimetron: [P2] Avoid treating every `not found` as a missing executable

This classifier is used for provider, model, session, and transport errors as well as spawn errors. A response such as `model ... not found` will therefore be labelled `Pi-Go could not start` and instruct the user to change `pi-go.command`, which will not fix the provider/model. Restrict this branch to ENOENT or command/binary-not-found forms, and cover the provider/model case with a test.

vscode/package.json
```diff
@@ -49,29 +50,27 @@
...
           "icon": "$(comment-discussion)",
           "contextualTitle": "pi-go"
         }
       ]
     },
     "commands": [
       {
-        "command": "pi-go.start",
-        "title": "pi-go: Ask (quick prompt)",
+        "command": "pi-go.openWalkthrough",
```
dimetron, line 67: [P2] Preserve or explicitly replace the quick-prompt entry point

This replaces the `pi-go.start` contribution, while the diff also removes its activation event and `extension.ts` handler. At the PR head, `vscode/README.md` still promises that `pi-go: Ask (quick prompt)` works without the proposed API, so the documented stable-activation path now points to a missing command. Restore the command/handler or update the documentation and provide an intentional migration path.

dimetron (commented):
— Codex review

vscode/src/ping.ts
dimetron: Fixed in 0c7a5b0. The formatter now detects unhealthy markers such as skipped endpoint probes and reports Ping as failed even when pi exits zero.

vscode/src/ping.ts
dimetron: Fixed in 0c7a5b0. Ping output now redacts sensitive URL query parameters, including key, token, authorization, and password values.

vscode/src/errorInfo.ts
dimetron: Fixed in 0c7a5b0. Launch details now redact secret-bearing flags and values before rendering them in error cards or native chat.

vscode/src/errorInfo.ts
dimetron: Fixed in 0c7a5b0. Missing-executable classification now relies on ENOENT and explicit spawn/executable wording; provider model-not-found errors remain request failures.

vscode/package.json
dimetron: Fixed in 0c7a5b0. README now documents the dedicated Pi-Go chat as the stable fallback instead of the removed quick-prompt command.

PR #352, opened 2026-09-18
feat(xai): use a SuperGrok subscription instead of per-token billing

dimetron:
## Why

xAI sells two products that both reach Grok, and pi-go could only use one of them:

- **A developer API key** (`XAI_API_KEY=xai-…`) — per-token billing against prepaid credits on `api.x.ai`.
- **A SuperGrok / X Premium+ subscription** — a weekly usage pool shared across Grok products.

A subscription already entitles the account to inference, but pi-go had no way to reach it. Sending a subscription bearer to the developer API is not merely unoptimized — it is rejected outright:

```
HTTP 402 {"code":"personal-team-blocked:spending-limit",
          "error":"You have run out of credits or need a Grok subscription."}
```

That error points at the subscription the caller already has, which makes it a confusing thing to be told. The subscription surface is a different endpoint: the Grok CLI chat proxy, which accepts an OAuth access token minted by `auth.x.ai` and gates it behind a set of Grok-CLI identity headers.

## What changed

**Credential handling — `internal/auth/xai.go` (new)**

- OAuth provider `xai` on the public Grok-CLI client, using the RFC 8628 device-code flow (no localhost listener, so it works over SSH and in containers).
- **Refresh grant.** xAI access tokens live six hours (`expires_in: 21600`). Without refresh, a login would stop working the same afternoon — not a usable arrangement for an agent harness. The response rotates the refresh token, so the rotated value is persisted; refresh is serialized because two concurrent refreshes would otherwise spend each other's token.
- Token store at `~/.pi-go/xai_auth.json` (0600, atomic temp+rename) carrying the refresh token and expiry, which the flat `~/.pi-go/.env` cannot express.
- Import of an existing `grok login` session, so a user who authenticated with the official CLI is not asked to do it twice.

**Routing — `internal/provider/xai.go`**

- A subscription token is detected structurally (a JWT carrying the xAI issuer or a `grok-cli:access` scope) and sent to the CLI proxy with the identity headers the proxy requires.
- An explicit base URL stays authoritative, so a gateway or test server is never silently replaced.
- A developer API key is never second-guessed by refresh.
- A provider may hold its credential outside its env var, so the CLI, the ACP runtime and the `/login` status consult the store before reporting `xai` as unconfigured.

**Command surface** — `pi login xai`, `pi logout xai`, and `/login` status.

## Two bugs found by testing rather than by reading

1. **The device-flow `OnToken` hook was missing** — the exact path `pi login xai` takes. Login would have reported success while never persisting a refresh token. `TestDeviceFlowInvokesOnToken` covers it; I confirmed it fails without the hook and passes with it.
2. **The CLI gated on `XAI_API_KEY` before the resolver ran**, so an expired token in `.env` — the real state six hours after login — was rejected rather than refreshed.

## Verification

| Gate | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./internal/...` | clean |
| `golangci-lint` (auth, provider, cli, tui, acp) | 0 issues |
| `go test ./internal/auth/` | pass |
| `go test ./internal/provider/` | pass |
| `go test -tags e2e ./internal/auth/` | pass |

Verified against live xAI before any code was written:

| Probe | Result |
|---|---|
| subscription bearer → `api.x.ai/v1/chat/completions` | `402 personal-team-blocked` |
| subscription bearer → `cli-chat-proxy.grok.com/v1/chat/completions` | `200` |
| refresh grant → `auth.x.ai/oauth2/token` | `200`, rotates, `expires_in: 21600` |
| device-code request | `200` |
| function calling via the proxy | returns `function_call` |
| token rotation, effect on the `grok` CLI's own token | non-destructive |

End-to-end, on a real binary with `XAI_API_KEY` unset and an **expired** stored token: the token refreshed, rotated, persisted, and a multi-turn tool-calling loop completed.

Two failures in the wider suite are pre-existing on `main` and unrelated — `internal/cli` `TestExecute` hangs, and `internal/acp/server` `TestNewPromptHandler_NoAPIKey` trips on a real `ANTHROPIC_API_KEY` in the environment.

## Caveat

xAI gates subscription API access per account and has been observed rejecting some tiers with 402/403 even on an active subscription. It works on the account this was developed against. The API-key path remains the fallback, and the README says so explicitly.


dimetron (commented):
Superseded by the consolidated final review submitted for this task.

dimetron (commented):
Superseded by the consolidated final review submitted for this task.

dimetron (commented):
Review of git diff main...HEAD: four actionable correctness findings.

— Codex review

internal/auth/xai.go
```diff
@@ -0,0 +1,537 @@
...
+}
+
+// refreshXAITokenLocked refreshes under a process-wide mutex and persists the
+// rotated set. The lock matters because xAI rotates the refresh token: a
+// concurrent second refresh would spend a token that is no longer valid.
+func refreshXAITokenLocked(ctx context.Context, refreshToken string) (*XAITokenSet, error) {
+	xaiRefreshMu.Lock()
+	defer xaiRefreshMu.Unlock()
+
+	set, err := RefreshXAIToken(ctx, refreshToken)
```
dimetron, line 432: [P1] Recheck the token after acquiring the refresh mutex. Both callers can pass the stale checks at lines 391–396, then the second waits on this mutex but still uses the refresh token it captured before waiting. After the first caller persists the rotated token, the second submits the already-spent token and can invalidate/fail the session. Reload the store and return the winner's fresh set (or otherwise coalesce refreshes) while holding the lock before calling RefreshXAIToken.

internal/auth/xai.go
```diff
@@ -0,0 +1,537 @@
...
+		return fmt.Errorf("setting token file permissions: %w", err)
+	}
+	if _, err := tmp.Write(body); err != nil {
+		_ = tmp.Close()
+		return fmt.Errorf("writing token file: %w", err)
+	}
+	if err := tmp.Close(); err != nil {
+		return fmt.Errorf("closing token file: %w", err)
+	}
+	if err := os.Rename(tmpName, path); err != nil {
```
dimetron, line 158: [P1] Make the atomic install replace-safe on Windows. On Windows, os.Rename returns an error when the destination already exists, so the first refresh/login can create xai_auth.json but every subsequent rotation fails at this line. That breaks repeat login and loses refresh-token persistence on the platform this PR explicitly made test-safe. Use a Windows-compatible atomic replacement strategy.

internal/cli/logout.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+}
+
+func runLogout(_ *cobra.Command, args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("a provider is required — supported: xai")
+	}
+	switch args[0] {
+	case "xai":
+		if err := auth.ClearXAITokens(); err != nil {
```
dimetron, line 33: [P1] Remove the mirrored subscription token from ~/.pi-go/.env during logout. saveResult/saveAdoptedXAI write the OAuth access token as XAI_API_KEY, but this command only deletes xai_auth.json; a later process therefore still authenticates with the leftover OAuth token. Preserve the entry only when it is a separate xai-* developer key, or stop mirroring subscription tokens into .env.

internal/cli/login.go
```diff
@@ -55,7 +62,22 @@ func runLogin(cmd *cobra.Command, args []string) error {
...
+	}
+
+	// A user who already ran `grok login` holds a usable subscription
+	// credential. Adopt it rather than making them authenticate twice; fall
+	// through to the interactive flow when there is nothing to adopt.
+	if prov.Name == "xai" {
+		if adopted, err := auth.ImportGrokCLITokens(); err != nil {
+			fmt.Printf("Note: could not read grok CLI credentials: %v\n", err)
+		} else if adopted != nil {
+			return saveAdoptedXAI(adopted)
```
dimetron, line 79: [P2] Preserve --model handling on the grok-session adoption path. Returning from saveAdoptedXAI here skips the flagLoginModel/config.SaveDefaultRole block below, so `pi login xai --model grok-*` silently fails to set the requested default while the normal OAuth path honors it. Apply the same post-login model handling before returning.

PR #358, opened 2026-09-20
fix(ollama): relax the think field when the model cannot think

dimetron:
## What

A session-wide thinking level of `"high"` was sent as `think:"high"` on **every** Ollama request, whatever model was selected. Ollama answers 400 for any model whose capabilities carry no `thinking` flag, and rejects the request *before* inference starts:

```
> explain project
✖ Error: 400 Bad Request: "qwen2.5-coder:7b" does not support thinking
```

The only guard was a substring match on the model name (`nothink`), which misses every non-thinking model not literally named that way — `qwen2.5-coder`, the embedding models, any older tag.

## How

Drop the `think` field on the rejection and retry **once**.

The thinking level is a *session* setting, not something the user asked this particular model for, so omitting the field is the correct request rather than a compromise. A wider name list was rejected as an approach because it goes stale and never covers third-party models.

- `ollamaThinkingUnsupported` — matches only that specific 400; an unrelated 400 still passes straight through.
- `ollamaRunStreamingRelaxThink` — wraps `retryStream`, relaxing only when **nothing has been forwarded yet**, so it cannot duplicate partial output.
- `ollamaRunNonStreamingRelaxThink` — same for the non-streaming path, where the failure arrives as a Go error rather than an `ErrorCode`.

The rejection is classified terminal by `retry.IsTransient`, so the retry follows in one round trip with **no backoff sleep** — the relax is not competing with the transient-retry budget.

The existing name-based `nothink` guard stays as a fast path.

## Verification

| Gate | Command | Result |
|---|---|---|
| Build | `go build ./...` | ✅ |
| Vet | `go vet ./internal/provider/` | ✅ |
| Lint | `make lint` | ✅ `0 issues.` |
| New tests | `go test -run 'TestOllamaRelaxThink\|TestOllamaThinkingUnsupported'` | ✅ 12/12 |
| Coverage | `-coverprofile` | ✅ 100% on all 4 functions, 96% package |
| Unit suite | `go test -p 1 ./...` | ✅ 65 ok / 5 no-tests / 1 pre-existing (see below) |

**End-to-end, real binaries against a mock daemon that reproduces the rejection:**

```
BEFORE (main):  EXIT=1  error: agent run: 400 Bad Request: "fake-coder:7b" does not support thinking
AFTER  (this):  EXIT=0  The project is pi-go.
```

Also verified against a live Ollama daemon: the error changes from `does not support thinking` to that model's own subsequent failure, proving the retry ran with `think=false`.

**The tests genuinely catch the bug** — reverting the two call sites fails 5 of them, one with the original error string.

## Pre-existing failures, unrelated to this change

Both confirmed by stashing this work and re-running:

- `internal/acp/server` `TestNewPromptHandler_NoAPIKey` — hits live Anthropic, 404 for the retired `claude-3-5-haiku-latest`. **FAIL** both ways.
- `internal/cli` `TestExecute` — **hangs indefinitely**, never panics, runs past its own 10m timeout. Excluded from the suite run above. Worth a separate look.

## Notes

- Scope is `internal/provider` only; no public API change.
- The local daemon on the reporting machine auto-updated (Ollama 0.33.3 → 0.34.2) mid-investigation and the originally-named model was removed from it, so the live repro used embedding models plus a mock rather than that exact tag.


dimetron (commented):
No actionable findings in `git diff main...HEAD`. `go test ./internal/provider`, `go vet ./internal/provider`, and `git diff --check main...HEAD` pass.

— Codex review

dimetron:
## `govulncheck` — no actionable findings

Run after the review was resolved, per repo policy (it reports on what the branch *depends on*, so it belongs after the review rather than before).

```

10 finding(s) with no fix available — reported, not failing
```

**Toolchain of record** — the answer is only true for the database on the day it ran:

| | |
|---|---|
| Go | `go1.27.1 darwin/arm64` |
| Scanner | `govulncheck@v1.8.0` |
| DB | `https://vuln.go.dev` (updated 2026-09-16 18:00:43 UTC) |

All 10 findings have **no released fix**, so the gate prints them and does not fail — there is nothing to upgrade to:

| ID | Module |
|---|---|
| GO-2025-3557, GO-2025-3558, GO-2025-3559, GO-2025-3582, GO-2025-3689, GO-2025-3695, GO-2025-3824, GO-2025-4251, GO-2026-5750 | `github.com/ollama/ollama` |
| GO-2026-5932 | `golang.org/x/crypto` |

None are introduced by this PR — it adds no dependency and changes no `go.mod`. The `ollama` module findings predate it: this branch only changes how pi-go *uses* an already-vendored client (`api.ChatRequest.Think`), and the dependency was already there.

`make check-cve` was deliberately **not** used: it opens with `go mod tidy -v`, which rewrites tracked files, and a check should not mutate the tree it is checking.

PR #360, opened 2026-09-20
feat(pimodels,piagent): expose common extension points

dimetron:
Exposes extension points that pi-go already holds internally. Two packages, both cases of a seam that exists but is unreachable from outside the module.

## `pimodels` — 4 of 6 unreachable transport options

`pimodels` left 6 of `provider.LLMOptions`' 15 fields unreachable, because it does not re-export the struct. Four were pure plumbing:

| New option | What it does |
|---|---|
| `WithMaxOutputTokens(n)` | Cap a reply, in tokens |
| `WithLegacyMaxTokens()` | Send `max_tokens` instead of `max_completion_tokens` |
| `WithXAITools()` | xAI server-side tools |
| `WithSystemCAsDisabled()` | Trust only the `WithCACert` bundle |

The remaining two (`TraceHTTP`, `TraceSink`) belong to PR #359, not here.

**`WithLegacyMaxTokens` is a correctness fix, not a preference.** Ollama understands only `max_tokens`; the newer `max_completion_tokens` is ignored by it *without an error*, leaving generation unbounded rather than rejected. An embedder pointed at Ollama — directly or through a gateway that does not rewrite the payload — previously had no way to set this.

## `piagent` — 2 seams `autocompact.Deps` already had

`internal/autocompact` exports `SummarizerLLM` and `Notify` as injectable Deps fields. `piagent` hardcoded both.

| New option | What it does |
|---|---|
| `WithSummarizer(m)` | Summarize with a different model than the one holding the conversation |
| `WithCompactNotify(fn)` | Be told when history is discarded, or when compaction fails |
| `Agent.LogPath()` | Find the session log |

`LogPath` closes a gap in *reasoning* rather than in code. `Notify` was `nil` on the grounds that "an embedder reads the outcome in the session log" (`piagent/agent.go:186-188`) — but `internal/logger` is under `internal/`, and there was no accessor, so the outcome was **discarded rather than relocated**.

## No behaviour change for existing callers

The summarizer still defaults to the session model, and the notifier stays absent unless asked for. Both are pinned by tests, including a control asserting `WithSummarizer` does **not** retarget `Agent.Model()`.

## Verification

- `go build ./...`, `go vet`, `gofmt` — clean
- `golangci-lint` on both packages — 0 issues
- `pimodels`, `piagent`, `internal/provider`, `internal/autocompact`, `internal/extension` — pass
- **Isolation tests** (`TestPimodelsStaysIsolated`, `TestPiagentStaysIsolated`) — pass; neither boundary moved
- **End-to-end from an external module** (`replace` to this worktree): a real `piagent.New` with `WithSummarizer` + `WithCompactNotify` + `WithContextWindow`, session created, turn completed, `LogPath()` resolving to a real file
- **Falsification**: breaking `WithLegacyMaxTokens` so it no longer sets the field makes the test fail with the intended message — the test is not vacuous
- 65 of 66 packages pass; the one failure is pre-existing on `main` (see below)
- Commit is signed; `git verify-commit` reports VALID

## Pre-existing failure, not from this branch

`internal/acp/server` `TestNewPromptHandler_NoAPIKey` — reproduced on unmodified `main`. It expects `"no API key"` but gets a live `404` from `api.anthropic.com` for `claude-3-5-haiku-latest`, so it depends on network and provider state.

## Not in this PR

`AddTools` (runtime tool registration) is the one genuinely missing capability — tools are frozen in `agent.New` and `Agent.Tools()` is read-only. TS's harness uses `registerTool` in 20 of its 84 extension examples. It needs design work (mutating ADK's tool slice under a lock, or rebuilding the `llmagent` and invalidating session wiring), so it is deliberately left for its own change.


dimetron:
## Codex review — resolved

Reviewed via `codex exec -s read-only review --base origin/main` (codex-cli 0.155.1). Two P2 findings, **both verified against the code and accepted**. A third defect in the same area was found while verifying them.

### ✅ [P2] `WithXAITools` had no observable effect — fixed in `6cfb5dc`

Confirmed: `provider.NewLLM` sets `opts.EnableXAITools = true` unconditionally for every `xai` model (`internal/provider/provider.go:704`), so an opt-in option could never see a false start.

Took the **remove** branch of Codex's suggestion, not the rewire. xAI server-side tools are on by default with `PI_NO_XAI_TOOLS` as the kill switch (`internal/provider/xai_tools.go:16`) — a *kill switch* is the control that was missing, and an opt-in was never it. Adding `WithXAIToolsDisabled` would be a separate decision about exposing env-var behavior as an option; the no-op option could not stay either way.

### ✅ [P2] Output caps ignored for native Ollama — fixed in `6cfb5dc`

Confirmed, and **the same defect applied to the sibling option**. Native Ollama (`internal/provider/ollama.go`) reads *neither* `UseLegacyMaxTokens` nor `MaxOutputTokens` — it caps via `PI_OLLAMA_NUM_PREDICT` — because it speaks Ollama's own API rather than the OpenAI wire. Codex flagged `MaxOutputTokens`; `WithLegacyMaxTokens` was documented next to it in the same broken example.

Rather than plumb a new cap into the native client (a behavior change to a shipping path, out of scope for this PR), the docs now:

- name the routes that *do* read both options (the OpenAI-compatible paths)
- show a gateway-based example that actually works
- state the `PI_OLLAMA_NUM_PREDICT` env var for the native route instead of implying an option reaches it
- carry a "which paths this applies to" note on each option's own doc comment

### ➕ Third defect, found while verifying

Not flagged by the review, same root cause: the README example paired both token options with `ollama/<name>`, so **neither** had an effect. Fixed together with the above.

### Regression tests added

| Test | Pins |
|---|---|
| `TestTokenOptionsDoNotReachNativeOllama` | `ollama/` routes native, not custom — so the README section stays true |
| `TestCustomBaseURLSelectsTheOpenAICompatibleClient` | the positive half: a base URL selects the client that *does* read both |
| `TestTransportOptionsOffByDefault` | no new option is on by default |

Routing assertions were **falsified** — inverting them makes the test fail with the intended message, so they are not vacuous.

### Verification after the fix

| Check | Result |
|---|---|
| `go build ./...` | ✅ |
| `go vet`, `gofmt` | ✅ clean |
| `golangci-lint` (both packages) | ✅ 0 issues |
| `pimodels`, `piagent`, `internal/provider`, `internal/autocompact` | ✅ pass |
| Isolation tests | ✅ boundaries unchanged |
| Both commits signed (`git verify-commit`) | ✅ VALID |

Note: the net option count is 12, not 13 — `WithXAITools` was removed.

— resolved in `6cfb5dc`

dimetron:
## govulncheck — pass

```
make vulncheck
```

| | |
|---|---|
| Scanner | `govulncheck@v1.8.0` (Go 1.27.1) |
| DB | `https://vuln.go.dev`, updated 2026-09-16 |
| Gate exit | **0 — pass** |
| Findings with a fix | **0** |
| Findings without a fix | 10 (reported, non-failing) |

```
10 finding(s) with no fix available — reported, not failing:
  GO-2025-3557  github.com/ollama/ollama
  GO-2025-3558  github.com/ollama/ollama
  GO-2025-3559  github.com/ollama/ollama
  GO-2025-3582  github.com/ollama/ollama
  GO-2025-3689  github.com/ollama/ollama
  GO-2025-3695  github.com/ollama/ollama
  GO-2025-3824  github.com/ollama/ollama
  GO-2025-4251  github.com/ollama/ollama
  GO-2026-5750  github.com/ollama/ollama
  GO-2026-5932  golang.org/x/crypto

```

**Nothing to upgrade here.** The gate fails only on findings naming a fixed version; every finding above has none released, so there is nothing actionable. This branch also adds no dependencies — it only wires options onto existing `provider.LLMOptions` fields — so the result is unchanged from `main`.

The answer is only true for the database on the day it ran: the DB was current as of 2026-09-16, so a re-run after the next DB refresh may differ.

dimetron:
Merged `origin/main` (`c9fe431`) to clear the conflict. The branch was `CONFLICTING`/`DIRTY`; it is now `MERGEABLE` with all 13 checks green on the merge commit `ff1e005`.

## The conflict

One file, `pimodels/README.md` — and it was purely additive on both sides:

- **This branch** added the Options table rows for `WithMaxOutputTokens`, `WithLegacyMaxTokens`, `WithSystemCAsDisabled`, plus the *Token limits on Ollama-backed endpoints* and *Trusting only one CA* sections.
- **`main`** (via #359) added the `WithTraceSink` table row and the *Tracing the traffic* section.

Nothing was contested, so I resolved it as a union: the `WithTraceSink` row appended to the table, and the *Tracing the traffic* section placed after *Trusting only one CA*. Both options keep their documentation.

`pimodels/pimodels.go` auto-merged cleanly — the two sides touched different regions (new `TraceEntry`/`WithTraceSink` declarations vs. the transport options).

## Verification

- `go build ./...`, `go vet ./...` — clean
- `gofmt -l` on changed packages — clean
- `pimodels`, `piagent`, `internal/provider` tests — pass
- The isolation and trace guardrails specifically re-run and passing: `TestTransportOptionsOffByDefault`, `TestTraceSinkCapturesWithoutGlobalEnable`, `TestTraceSinkIsPerClient`, `TestNoTraceSinkCapturesNothing`, `TestTraceEntryFieldsReachable`, `TestTraceMaxBodyMatchesTransport`
- Section-by-section and table-row diff against both parents confirms **no content was dropped from either side**
- Merge commit is signed (`git verify-commit` reports VALID) and carries a `Signed-off-by` trailer, per AGENTS.md
- CI: 13/13 pass on `ff1e005`

## Remaining blocker

`mergeStateStatus` is `BLOCKED` by `REVIEW_REQUIRED`, not by the conflict — the branch needs an approving review. That is a human gate, so this is ready for review rather than for merging.

PR #362, opened 2026-09-20
docs(specs): compare pi-go against DevoxxGenie and rank features to adopt

dimetron:
A source-level comparison of pi-go against the DevoxxGenie IntelliJ plugin (v1.14.2), producing a ranked list of features worth adopting.

## What this is

Documentation only — `specs/issues/compare-devoxxgenie.md`, 403 lines. No code changed in either repository.

The two projects are different species, and most of the document follows from that: DevoxxGenie is IDE-native (Java, IntelliJ Platform, Langchain4j), pi-go is a terminal agent harness (Go, ADK). pi-go leads on agent runtime; DevoxxGenie leads on human control of the agent and editor-native surfaces.

## Recommendations

| # | Action | Effort | Impact |
|---|--------|--------|--------|
| 1 | `before_tool` hook denial + destructive-command matcher | small | **highest** — closes the largest safety gap |
| 2 | Per-turn tool/turn budget in `internal/agent`; move stuck detector out of TUI | small–med | prevents unbounded runs in headless/CI |
| 3 | Session cost accumulation + `/cost` + status bar | small | makes an existing guardrail legible |
| 4 | `ToolCallContentDiff` over ACP | small | three editors, and deletes duplicated logic |
| 5 | Wire Palace tools + `WakeUp` into TUI and ACP server | small | fixes an unreachable headline feature |
| 6 | Bounded repo map | medium | judgment call — may fight context tooling |

The top three are safety and visibility gaps rather than missing features.

## pi-go defects found

1. **The Memory Palace is unreachable from the TUI and ACP server** — zero references to the package in either; `WakeUp` is called only at `cli.go:1316`.
2. **Subagent and skill `tools:` frontmatter is parsed and never enforced** — `agents.go:36,144` sets it, `spawner.go:124-128` forwards only `--system` and `--lsp`.
3. **The `grep` tool picks its own name at build time** (`grep.go:131-134`), so anything naming it by string is host-dependent.
4. **Cost is fully implemented and never displayed** — `provider.CostFor` has exactly one non-test caller, and `pirpc` hardcodes `cost: 0`.

## Verification

Every load-bearing claim carries a `file:line` plus the re-verification command used, in the method section, so each can be checked in one step. The 41-tool inventory in the appendix was dumped by executing `eval.Inventory()` rather than transcribed from source. Doc/code disagreements in the plugin are recorded separately, because they mean its documentation is not a reliable design source.

## Note on scope

This PR contains the comparison document only. It was rebased onto `origin/main` to exclude 15 unrelated unpushed commits (an in-progress `026-async-input-subagent` spec, plus changes to `internal/provider/ollama.go`) that were sitting on local `main`; those remain on local `main` and its backup branch, untouched, since there is an active worktree for that work.


dimetron:
Tracking the findings and recommendations from this document as issue #364.

The four pi-go defects and six ranked recommendations now have an issue to discuss and schedule them against, since this PR is documentation only and does not implement any of them.

Splitting the defects into individual issues is possible if preferred — noted in #364.

PR #365, opened 2026-09-20
specs(010): SoL-Pi efficiency lessons — archive-before-reduce plan

dimetron:
## What this is

A review document, not an implementation. It records a study of `tmp/SoL-Pi` — NVIDIA's standalone extension for the Pi (TypeScript) agent harness — and turns the transferable findings into a scoped spec with an implementation plan for five unblocked items.

**No Go code changed. No build or test was run.** Every `file:line` anchor in `research.md` was read from the tree; claims that are inferred rather than read are marked `(inferred)`.

## The finding

pi-go is **not behind** SoL-Pi on capability. It already ships mechanisms SoL-Pi has no equivalent for:

| Capability | pi-go | SoL-Pi |
|---|---|---|
| Deterministic log compactor (no model call) | ✅ `internal/tools/compactor*.go` | ❌ |
| Byte-identical result deduper | ✅ `internal/tools/dedup.go` | ❌ |
| Superseded-result shedding | ✅ `internal/session/compaction_shed.go` | ❌ |
| Window resolution from model catalog | ✅ `internal/ctxwindow` | ❌ |

What pi-go lacks is two **architectural commitments**:

1. **A separation between durable history and the model-visible projection.** pi-go collapses them, then destroys the record to shrink the view: `rewriteEvents` (`internal/session/store.go:1291`) truncates `events.jsonl` in place. The shed stub tells the model to re-run a command to recover bytes pi-go already had.
2. **A reduction that can be rejected rather than a mutation already applied.** `applyCompaction` (`internal/tools/compactor.go:119`) mutates the result map in place with no rejection point and no retained original.

```mermaid
flowchart TD
    subgraph now["Current — one layer"]
        T1["tool result"] --> TR1["truncateOutput<br/>discards past 256KiB"] --> C1["compactor<br/>mutates in place"] --> E1["events.jsonl<br/>rewritten in place"] --> P1["provider"]
    end
    subgraph next["Target — two layers"]
        T2["tool result"] --> AR["archive (redacted)<br/>+ readback pointer"] --> C2["compactor<br/>returns applied bool"] --> E2["events.jsonl<br/>keeps pointer"] --> P2["provider"]
        AR --> ART["tool-output/&lt;id&gt;.txt<br/>durable"]
        ART -.->|"read offset=N"| M["model asks again"]
    end
```

## Corrections that change the implementation

Both are flagged `[CORRECTION]` in `research.md`. An earlier verbal analysis of mine was wrong twice, and implementing from the summary rather than `research.md` would produce a broken change.

**1. It is not one dead stage, and not four — it is seven of nine.** Three independent defects compose, and only fixing all three revives a route:

- *Field mismatch.* `compactGrep` reads `result["output"]` (`compactor_search.go:10`) but `GrepOutput` (`grep.go:112-119`) emits `matches`/`total_matches` and **no `output`**; same for `compactFind` (`files`) and `compactGitOverview`. Returns `nil` unconditionally.
- *Name mismatch.* Three git pipelines route on underscores (`compactor.go:108-113`) while the tools register hyphens (`git_diff.go:36`).
- *No route at all.* `ls` is registered (`registry.go:66`) with no `case` in `compactToolResult`.
- *And `applyCompaction` (`compactor.go:119-156`) is a third defect*, not a consequence of the above: fed a hand-made non-nil `*CompactResult`, it still declines to write for grep/find/tree/git-overview and logs `no known output field`, because it probes only `stdout`→`content`→`output`→`diff`→`result`→`data`.

Verified by running the real `BuildCompactorCallback` against real production result shapes: **2 of 9 registered bulk-output tools compact (`bash`, `read`); 7 do not.** `CompactMetrics.Summary()` after a full run reports `ByTool` for `bash` and `read` only.

An earlier draft of this body said "four dead stages" and the first rewrite of `research.md` said "seven of nine" on evidence that was wrong in two opposite, cancelling ways — the probe listed `grep` **and** `ripgrep` as separate routes when `grep.go:128-133` registers exactly one of them, and it omitted `ls` entirely (`9 − 1 duplicate + 1 omission = 9`, which is why the fraction survived). The corrected set of seven is `ripgrep`, `find`, `tree`, `ls`, `git-file-diff`, `git-overview`, `git-hunk`. **The denominator is now sourced from the registry, not a hand-written list** — that is what let both errors through.

**2. Archiving before truncation would write secrets to disk in cleartext.** `budgetStreams` (`bash_supervisor.go:440-443`) applies `redactSecrets` **after** `truncateOutput`, so truncation's input — what an archive-before-truncate design persists — is unredacted. Decision **D2** exists solely for this, and T1's most important test seeds a payload with `sk-…`, a JWT and an `AKIA…` key and asserts the artifact on disk contains `***`.

Also corrected during research: `read` addresses by **line**, not byte (`read.go:141`), so the readback pointer is line-based; and session-dir paths are **blocked** by the sandbox (`sandbox.go:59-72`), so the artifact root needs `AddExtraDir`.

## Scope — five unblocked items

| # | Item | Blocked by | Risk |
|---|---|---|---|
| T1 | Archive-before-reduce + readback pointer | nothing | medium |
| T2 | Compactor dispatch (7 dead stages; `ls` gets a route) | nothing | low |
| T3 | Make reduction rejectable | nothing | low |
| T4 | Wire the dead measurement surfaces | nothing | low |
| T5 | Ollama cache reporting | upstream | medium |

T1 and T2 are independent and can run in parallel; T3 needs both; T4 needs T1+T2 so it has something real to measure.

## Deliberately excluded

Recorded in `README.md` so the exclusions are on the record rather than merely absent:

- **Compaction economics gate** — blocked on T5. `Decide` (`compaction.go:123-137`) reads only `bodyTokens`/`windowSize`; a price gate before cache numbers exist would price on wrong inputs for 73% of traffic.
- **Semantic / plan-phase triggers** — need T3 + T5, plus a correctness argument for cutting mid-run. pi-go's turn-level constraint is legitimate (`agent.go:377-381`).
- **Action Fusion** — rejected on the merits, not deferred. pi-go's own measurement: 81.8% of tool-issuing turns make exactly one call, and batching (48–65%) beats per-call fusion.
- **`pi tokens`, OTel metrics, cost computation** — owned by `specs/issues/token-cost/07-measurement.md`.

## Decision needed before `PROMPT.md`

Six decisions are left open in `design.md`, each with a recommendation. D2 (archive before or after redaction) is the one not to proceed without.

## Relationship to existing work

`specs/issues/token-cost/` is the authoritative measurement effort (1,404 sessions, 2.5B prompt tokens, 201:1 ratio). This spec references it rather than re-deriving it. Two overlaps are by reference: item 5 expands `token-cost/02-cache-reporting.md` with code anchors; item 4 overlaps `token-cost/07-measurement.md` rec 3 but is scoped to the wired-but-dead surfaces only.

## Verification

- `golangci-lint run ./cmd/... ./internal/... ./hack/...` → `0 issues` (run before commit).
- Commit signed and verified with `git verify-commit` (not a header check): `Good "git" signature for dimetron@me.com`.
- `Signed-off-by` trailer present.

- **R2 re-verified in a second, independent pass:** the real `BuildCompactorCallback` re-run against real production output structs with `ls` included, plus a live `CoreTools` call to enumerate the registered names. That pass is what found the `grep`/`ripgrep` double-count and the missing `ls` row; its `ls` byte count comes from its own payload and `research.md` §R2.4 says so. Every claim it added — registered names, the `ls` verdict, the `applyCompaction` failure, the `git-file-diff` asymmetry — was measured, not inferred.

No functional change, so no build or test result is claimed. The temporary probes were removed; `git status --short` is empty and no `.go` file is in the diff.


dimetron:
## Update — PROMPT.md added, and the compactor finding was understated

`PROMPT.md` is now in the branch. While writing it I verified the compactor claim
empirically instead of by reading, and **the first commit understated it**. Pushed
as `20e906e`.

### Seven of nine compaction pipelines are no-ops in production

I fed real production result structs through the real `BuildCompactorCallback`:

```
bash             COMPACTED      before= 11202 after=    92
read             COMPACTED      before= 11202 after=    92
ripgrep          DEAD (no-op)   before= 25091 after= 25091
grep             DEAD (no-op)   before= 25091 after= 25091
find             DEAD (no-op)   before=  7201 after=  7201
tree             DEAD (no-op)   before= 11202 after= 11202
git-file-diff    DEAD (no-op)   before= 11202 after= 11202
git-overview     DEAD (no-op)   before=  7201 after=  7201
git-hunk         DEAD (no-op)   before= 32401 after= 32401
```

**Cause.** ADK marshals each tool's typed output struct to JSON and unmarshals
into `map[string]any` (`adk/v2@v2.4.0/tool/functiontool/function.go:231`), so the
result keys are the structs' `json` tags. `compactGrep`, `compactFind` and
`compactGitOverview` read `result["output"]` — **a key no tool's output struct
produces**. They return `nil` unconditionally.

**Why the suite missed it.** `compactor_test.go` feeds these pipelines synthetic
maps with an `output` key (`:702,709,784,999,1019,1028,1061,1076`). The tests
assert against a shape no tool emits, so the pipelines were green while dead.

So the previous message's "four dead stages" was still wrong — it's seven of
nine, and the two that work are `bash` and `read`. The practical consequence:
the compactor's real contribution to pi-go's token traffic today is `bash` and
`read` only, and any expectation based on the other seven is unfounded.

### What changed in the branch

| File | Change |
|---|---|
| `PROMPT.md` | **new** — T1–T5 execution prompt |
| `research.md` | R2 rewritten: probe, output, real key sets per tool, five-part fix scope |
| `plan.md` | T2 rescoped into T2.0–T2.4 (was three line-level fixes) |
| `design.md` | Success criterion 2 tightened — test must use real structs |
| `README.md` | Notes the scope growth |

`PROMPT.md` opens by telling the implementing agent **not** to work from the plan
summary, because R2 changed after the plan was written. That is the failure mode
this correction exists to prevent.

### Verification

- Probe files were temporary and removed; tree is clean.
- `go build ./...` → OK. `go test ./internal/tools/` → `ok 27.858s`.
- `golangci-lint` (pre-commit) → `0 issues`.
- `git verify-commit HEAD` → `Good "git" signature for dimetron@me.com`.

Still no Go code changed. The six design decisions in `design.md` remain open —
D2 (archive before or after redaction) is the one not to proceed without.

PR #368, opened 2026-09-20
fix(tools): revive the 7 dead compaction pipelines, and disclose every cut

dimetron:
> **Correction, added after two reviews.** The savings table below measured the
> compaction callbacks **in isolation**. ADK runs only the FIRST after-tool
> callback that returns a non-nil result, and the OTEL tracing callback is
> registered first and always returns the result map — so dedup and the compactor
> **never ran in production**, and none of those percentages were reachable.
> That is now fixed in `5da126a` (the chain is composed via
> `extension.ComposeAfterToolChain`, verified with a real agent run); the numbers
> are kept because they are still the correct measurement of the pipelines
> themselves, but they describe a capability that only became live with that
> commit. See *The chain was composed, not the pipelines* below.

## What this fixes

Seven of the nine registered compaction pipelines were **no-ops in production**. They were green in tests and computed nothing at runtime.

Measured through the real `BuildCompactorCallback`, with each payload capped at the **bound the tool itself enforces** — a payload larger than the tool can emit measures a shape that never reaches the compactor:

| tool | source bound | before | after | saved |
|---|---|---|---|---|
| ripgrep | 200 matches (`grep.go:20`) | 21815 | 10832 | 50.3% |
| find | 500 files (`find.go:14`) | 13419 | 2636 | 80.4% |
| ls | 1000 entries (`ls.go:11`) | 47814 | 4631 | 90.3% |
| git-file-diff | 1500 changes, ≤256KB (`git_diff.go:87`) | 84844 | 2799 | 96.7% |
| git-hunk (240 small hunks) | unbounded (whole-file diff) | 21885 | 9145 | 58.2% |
| git-hunk (1 large hunk) | unbounded (whole-file diff) | 63987 | 26213 | 59.0% |
| tree | 256KB (`tree.go:87`) | 200035 | 11057 | 94.5% |
| bash | 256KB | 200039 | 11062 | 94.5% |
| read | 256KB | 200033 | 11056 | 94.5% |
| **git-overview** | **≤10 commits** (`git_overview.go:72`) | 257 | 257 | **no-op** |

**`git-overview` is a correction to my own earlier claim.** I first reported it saving 49–76%. Those figures came from feeding it 200 commits, but the tool runs `git log --oneline -10` — it can never return more than 10. The compactor's `MaxLogEntries` cap is 40, so that pipeline is **unreachable in production by construction**: wired and correct, but it can never fire. A test now records this (`TestCompact_NoOpBelowItsCap`) so the number is not claimed again.

Three independent root causes, all silent mismatches rather than crashes:

1. **`applyCompaction` guessed its target field.** It probed `stdout → content → output → diff → result → data` in order. No tool emits `output`, so a correct pipeline could still write to the wrong key — or to none, which is what produced `compactor: no known output field in result for replacement`. `CompactResult` now carries explicit `CompactWrite{Key, Value}` pairs, so a pipeline can only write the field it read.
2. **Routing used names the tools never register.** `compactToolResult` switched on `git_file_diff` / `git_overview` / `git_hunk` while the tools register hyphens (`git_diff.go:36`, `git_overview.go:52`, `git_hunk.go:42`) — three pipelines unreachable. It also matched only `grep`, but the search tool self-names `ripgrep` whenever `rg` is on PATH (`grep.go:128-133`, built once by `registry.go:62`). Routing is now a map keyed by registered name, so a test can enumerate it.
3. **Pipelines read fields their tools do not emit.** `compactGrep`/`compactFind` read `output` instead of `matches`/`files`; `compactGitOverview` read `output` instead of `recent_commits`; `compactGitHunk` read `diff`, a field `git-hunk` has no equivalent of; `compactTree` delegated to `compactFind`, so it died whenever `compactFind` did; and `ls` is registered (`registry.go:66`) with **no pipeline at all**.

## The spec's one wrong instruction

The spec said `compactGrep` must "read `matches` (a `[]GrepMatch`)". It cannot: ADK round-trips every result through `json.Marshal`/`json.Unmarshal` (`adk v2.4.0 internal/typeutil/convert.go`), so `matches` arrives as `[]any` of `map[string]any` — never a typed slice, and a typed assertion would fail.

More importantly, re-rendering the array to a string would have broken the TUI, which reads those same keys as `[]any` (`internal/tui/tool_display.go:897-984`). The caps therefore **trim the array and keep its type**.

## Information loss found while auditing the wins

A large saving is only a win if the agent can still act on what remains. Auditing for that found **two real defects in the compactor's own truncation**, both now fixed:

**1. An unbalanced per-hunk cut was silent and misleading.** The per-hunk cap keeps the first `MaxDiffHunkLines` change lines. When those lines are all one sign, the visible hunk shows only that sign — measured before the fix: a balanced 30-deletion/30-addition change whose cut landed inside the deletions rendered as 20 deletions and **0 additions**, with no note. The file tally `(+30 -30)` could not be reconciled with the visible lines, so the diff read as a pure removal. The compactor now emits `... (1 deletion, 1 addition truncated)`, split by sign — the same disclosure rtk makes (`git_cmd.rs`, `hunk_truncation_note`), for the same reason: an anchored `^-`/`^+` audit must be able to tell what it did not see.

**2. `git-file-diff` never set its own `truncated` field.** The tool declares it (`git_diff.go:32`) and sets it when its 256KB byte cap fires (`git_diff.go:88-90`), but the compactor could cut a diff by 99% and leave the result claiming to be the complete diff for that file. Now reported the same way.

Both were the same `omitempty` trap that also bit me earlier in this PR: `truncated` is absent from a complete result, so a write gated on the key already existing means the flag is **never** set. Writes for declared fields are now unconditional.

**Known limitation, documented not hidden:** a count-based cap keeps a *head* slice, not a spread. Capping 500 matches across 50 files leaves 100 matches covering only **10 distinct files** — the agent learns the true total (`total_matches`) and that the result is partial (`truncated`), but not which files were dropped. rtk addresses this by grouping per file and emitting a `+N more files` hint (`search.rs`); pi-go does not yet. There is a test recording the current coverage as a baseline rather than asserting it is correct.

## pi-go built-in vs rtk, on identical input bytes

rtk is available here (v0.49.0, `/opt/homebrew/bin/rtk`), so both sides were measured on the same fixtures. The architectures differ: rtk is a **CLI proxy that runs commands**, pi-go's compactor **post-processes a result it already has**.

| tool | raw | pi-go | saved | rtk | saved | rtk mode |
|---|---|---|---|---|---|---|
| grep (400 matches) | 28872 | 10022 | 75.5% | 28872 | **0.0%** | `pipe -f grep` — byte-identical |
| grep-big (3192 matches) | 270200 | 11551 | 96.8% | 17020 | 93.7% | `rtk grep -rn` (command) |
| find | 10690 | 2836 | 75.4% | 1055 | 90.1% | `pipe -f find` |
| ls | 23890 | 4840 | 75.2% | 797 | 96.7% | `rtk ls` (command) |
| tree | 15804 | 13800 | 15.7% | 15784 | 0.1% | `rtk tree` (command) |
| git-file-diff | 222891 | 966 | 99.6% | 3722 | 98.3% | `pipe -f git-diff` |
| git-log | 32603 | 2057 | 79.6% | 185 | 99.4% | `pipe -f git-log` |
| git-status (472 files) | 6719 | 7096 | 0.0% | 6717 | 0.0% | `pipe -f git-status` |
| git-hunk | 89260 | 43301 | 58.0% | n/a | — | rtk has no hunk filter |

**Where pi-go wins:** `git-file-diff` (99.6% vs 98.3%), and the cases rtk cannot do at all — `git-hunk` has no rtk filter, and the structured `grep`/`ls`/`tree` results are capped in place rather than re-parsed from text.

**Where rtk wins:** `find` and `ls` compact harder by grouping; `git-log` caps the commit list to 10.

**Where both correctly do nothing:** `git-status`. Neither truncates the dirty-file list, because hiding which files are dirty misleads about repo state. That 0.0% row is intended behaviour on both sides, not a defect.

Two rtk behaviours argue against delegating to it, which is why this PR ports its *decisions* rather than wiring in the binary:

- **`rtk pipe -f grep` was byte-identical** on a 400-match list — the same no-op class this PR fixes.
- **`rtk pipe -f go-test` lost a real failure**: on genuine `go test` output (`FAIL: TestBad (0.00s)` / `x_test.go:6: expected 1 got 2`) it emitted `Go test: No tests found` with **exit code 0**, discarding both the failure and its signal.

So the ported ideas are: caps that preserve the true pre-cap total, a never-worse guard, per-hunk truncation disclosure, and capping logs while leaving status lists whole.

## The chain was composed, not the pipelines

Both reviews independently found the same defect, and it is the most important
thing in this PR.

`Flow.invokeAfterToolCallbacks` (`adk/v2@v2.4.0 internal/llminternal/base_flow.go:1436`)
returns at the **first** callback that yields a non-nil result:

```go
for _, callback := range f.AfterToolCallbacks {
    result, err := callback(...)
    if result != nil { return result, nil }   // stops here
}
```

Every pi-go after-tool callback returns the result map, and OTEL tracing is
registered first and always returns it. So the chain stopped at tracing and
dedup, the compactor, the LSP after-hook and memory recording were all skipped.

**Measured, not inferred.** A real ADK agent wired with two after-tool callbacks
ran only the first, and the second's mutation never reached the model (25600
bytes arrived untruncated). With the stages composed into one callback, all ran
and the change landed (25600 → 335 bytes). `TestComposedAfterChainReachesCompactor`
pins that end-to-end, and the composed chain is asserted to be exactly one
callback at every wiring site.

The fix mostly existed already: `piagent/composeAfterTool` folded the chain with
passing tests, but only `piagent` used it. The logic moved to
`extension.ComposeAfterTool` / `ComposeAfterToolChain` — the package all four
sites already import — and `cli.go`, `interactive.go` and `acp/server/runtime.go`
now compose.

Two other findings from the reviews, both fixed:

- **`callback_order_test.go` modelled ADK wrongly.** It fed every callback's
  output forward regardless of nil-ness — the opposite of the short-circuit — so
  it passed while production stopped at the first entry. It now asserts a single
  composed callback and drives it with a real `agent.Context`.
- **`dedupTools` overstated coverage.** `git-hunk` (an array payload) and
  `git-overview` (no recognised key) can never dedup, and are removed. The test
  that replaced them exposed a wider truth: only `read`, `read_image` and
  `git-file-diff` can actually dedup — `grep`/`find`/`ls`/`tree` cannot either,
  because their payloads are arrays. Extending the probe to arrays is a feature
  change, so `TestDedup_ReachableFieldsAreHonest` pins today's behaviour and
  records the follow-up instead of assuming it.

`TestCompactGitOverview_CapsCommitsNotFileLists` also fed 200 commits — an input
the tool cannot emit (`git log --oneline -10`) and the exact anti-pattern this
PR's own AGENTS.md section calls out. The oversized fixture is now explicit about
why it exists and points at the unreachability test.

## Review findings addressed

A Codex review raised four actionable items. Each was verified against the code before acting; **three were real, and one needed a different fix than proposed**.

**1. Hunk notes were attributed to the wrong hunk.** `flushHunkNote` ran only at file boundaries and at the final cut, so a truncated hunk's withheld counts were carried into the *next* hunk. Measured: hunk 1 cut to 3 of 10 changes emitted its note **after hunk 2's header**, attributing one hunk's loss to another. The note is now flushed when a new `@@` header opens, so it sits beside the hunk it describes. Pinned by `TestCompact_HunkNoteStaysWithItsHunk`.

**2. Capping the number of hunks left `Content` unbounded.** A hunk is not limited by how many hunks exist — `git-hunk` can return **one** hunk with hundreds of KB, and `len(hunks)=1` never trips a count cap. Measured: a single **223890-byte** hunk passed through untouched. Each hunk's `Content` is now capped on a line boundary with the dropped line count disclosed, while the hunk's own `Added`/`Removed` counts stay exact. Pinned by `TestCompact_LargeHunkContentIsCapped`.

**3. The opt-out switches were dropped.** `compact_git_output` and `group_search_output` were honored by the pipelines before this fix; my new paths ignored both, which would leave a documented setting silently ineffective. Both are honored again. For `git-file-diff` the gate covers the semantic rewrite but **not** hard truncation — matching the previous behaviour, where the byte ceiling also ran unconditionally. Technique names now reflect what actually ran. Pinned by `TestCompact_RespectsConfigOptOuts`.

**4. The routing guard assumed `ripgrep` is registered.** `newGrepTool` registers `grep` when `rg` is absent (`grep.go:128-133`), so requiring `ripgrep` would fail the supported no-`rg` configuration — including Windows CI. The test now asserts **exactly one** of `grep`/`ripgrep` is registered, with both routing entries checked separately.

**Where I did not follow the review:** it asked to align `git-overview`'s `MaxLogEntries` cap with its data (or remove it). The tool fetches only 10 commits (`git_overview.go:72`), so the cap cannot fire. Rather than change the tool's behaviour, the unreachability is recorded as a test so the pipeline is not reported as a saving it never makes.

## Also in this PR

- **Removed two `log.Printf` calls from the compaction path.** They ran inside an `AfterToolCallback` while the TUI owns the alternate screen, where any stdout/stderr write corrupts the display (AGENTS.md, *TUI output safety*). The panic is now recorded as a technique, so it stays observable through metrics instead of stderr.
- **AGENTS.md** gains a *Tool-output compaction* section recording the invariants and the rtk decision with the measurements above, so this cannot silently regress.
- **AGENTS.md** also gains *Never trust a big win without a proof check*, the lesson this PR taught. It records the three defects that hid behind the savings numbers — an impossible input shape (`git-overview` measured at 200 commits when the tool caps at 10), synthetic fixtures carrying a key no tool emits, and a per-tool saving never proven end-to-end — plus the rules that follow: bound the input by what the producer can emit and state the bound, build fixtures from real output structs, prove the saving end-to-end rather than at the seam, assert the consumer's fields survive rather than only that output shrank, confirm a guard can fail by reintroducing the bug, run the cheapest falsification before a full eval, and label any unverified number. It points at the real harness (`make eval-tools`, `make eval-run`, `make eval-judge`) and names the concrete gap: `AvgResultBytes` is reported in every eval but never gated against a baseline, which is why seven dead pipelines looked healthy.

## Verification

- `go build ./...` → OK
- `make lint` → `0 issues`
- `go vet ./internal/tools/`, `gofmt -l` → clean
- `go test ./internal/tools/` → ok (all tests pass)
- Tests build payloads by marshalling the **real output structs** (`buildProdResult`). The previous tests fed synthetic maps carrying an `output` key that no tool emits — that is why the pipelines were green while dead.
- Guards verified to **fail when the bug is reintroduced**, then pass when restored: reintroducing the `git_hunk` underscore routing failed `TestCompactor_EveryToolCompacts` and `TestCompactorRouting_EveryRegisteredTool`; gating the `truncated` write on the key already existing failed the disclosure test. A guard that cannot fail is worthless.
- `compactor_preservation_test.go` pins the information-preservation contract: every tool discloses its cut and keeps its true total; the unbalanced hunk cut is disclosed; `git-overview` keeps dirty-file lists whole; a cap above a tool's own source cap cannot fire.

One pre-existing issue, unrelated to this change and verified on a clean base commit: `TestExecute` in `internal/cli` hangs (it blocks rather than failing). I confirmed it hangs identically at `b1a63ac`, so it is not caused by this PR and I have not touched it.

No Go code changes outside `internal/tools/compactor*` and AGENTS.md.


internal/tools/compactor_test.go
```diff
@@ -988,122 +1037,272 @@ func TestCompactRead_SmallContent(t *testing.T) {
...
+			continue
+		}
+		if _, ok := compactorPipelines[x.Name()]; !ok {
+			t.Errorf("registered tool %q has no compaction pipeline", x.Name())
+		}
+	}
+
+	for name := range wantPipelines {
+		if !registered[name] {
+			t.Errorf("expected %q among registered tools; registry changed shape", name)
```
dimetron, line 1302: [P1] Make this assertion tolerate hosts without `rg`. `newGrepTool` intentionally registers `grep` when `rgAvailable` is false, but `wantPipelines` always contains `ripgrep` and this loop requires every listed name to be registered. The Windows CI run fails here with `expected "ripgrep" among registered tools`, so the new guard breaks the supported no-rg configuration. Assert that exactly one of `grep`/`ripgrep` is registered (while checking both routing-map entries separately), or derive the expected name from the actual registry.

internal/tools/compactor_git.go
```diff
@@ -6,122 +6,117 @@ import (
...
 	}
 
-	origSize := len(diff)
-	var techniques []string
-
-	if cfg.CompactGitOutput {
-		diff = runStage(diff, &techniques, "git-compact", func(s string) (string, bool) {
-			return compactGitDiffText(s, cfg)
-		})
+	capped, cut := capArray(commits, cfg.MaxLogEntries)
```
dimetron, line 66: [P1] This cap cannot run for the production default. `GitOverviewOutput.RecentCommits` is populated by `git log --oneline -10` in `git_overview.go`, while `DefaultCompactorConfig.MaxLogEntries` is 40, so the real result can never exceed this cap and this pipeline remains a no-op. The test's 200 synthetic commits do not match the tool's production shape, which hides that regression. Either compact an actually oversized field, raise/change the upstream limit, or align the cap and test with the real output contract.

internal/tools/compactor_git.go
```diff
@@ -6,122 +6,117 @@ import (
...
+// The old versions read "diff" for git-hunk (a field it does not have) and
+// "output" for git-overview (a field no tool has), and routed on underscore
+// names the registry never produces — so all three were no-ops.
+
+// compactGitFileDiff compacts the diff text of a git-file-diff result.
+//
+// This pipeline already read the right field ("diff") and applyCompaction
+// already wrote it; only the name was wrong. It is therefore the smallest fix
+// in the set, and the one to land first.
+func compactGitFileDiff(result, _ map[string]any, cfg CompactorConfig) *CompactResult {
```
dimetron, line 32: [P2] Preserve `CompactGitOutput=false` for dedicated git tools. The previous implementations guarded their git-specific compaction stages with this flag; the new `git-file-diff`, `git-overview`, and `git-hunk` paths never inspect it, so disabling git compaction still caps or rewrites their results. This makes the documented setting ineffective for callback-routed git tools. Gate the new git-specific stages/caps on the flag and add a false-case test.

internal/tools/compactor_search.go
```diff
@@ -1,80 +1,152 @@
...
 	}
 }
 
-// compactFind applies hard truncation to find tool output.
-func compactFind(result map[string]any, cfg CompactorConfig) *CompactResult {
-	output, _ := result["output"].(string)
-	if output == "" {
+// compactGrep caps the matches array of a grep/ripgrep result.
+func compactGrep(result, _ map[string]any, cfg CompactorConfig) *CompactResult {
+	return capResult(result, "matches", "total_matches", "truncated", "search-cap", cfg.MaxSearchTotal)
```
dimetron, line 110: [P2] Preserve the `GroupSearchOutput`/per-file limit semantics. The old grep pipeline only grouped and limited matches when `cfg.GroupSearchOutput` was true; with it false it fell back to generic hard truncation. This new path always caps the structured array and ignores both `GroupSearchOutput` and `MaxSearchPerFile`, so the documented false-toggle behavior and per-file limit are lost. Keep structured compaction behind the flag or implement the fallback semantics, with tests for both modes.

dimetron (commented):
The focused tests pass locally, but the PR still has one cross-platform test failure and several compaction paths that do not match the production result/config semantics. Details are inline.

— Codex review

internal/tools/compactor_git.go
```diff
@@ -6,125 +6,158 @@ import (
...
+// in the set, and the one to land first.
+//
+// It sets the tool's own `truncated` field when the diff is cut. Before this,
+// the compactor could reduce a diff by 99% while leaving truncated unset, so the
+// result claimed to be the complete diff of the file. GitFileDiffOutput already
+// declares the field (git_diff.go:32) and the tool sets it when its own 256KB
+// byte cap fires (git_diff.go:88-90) — compaction is the same kind of loss and
+// must report it the same way.
+func compactGitFileDiff(result, _ map[string]any, cfg CompactorConfig) *CompactResult {
+	return gitTextResult(result, "diff", cfg,
```
dimetron, line 33: [P1] Preserve the `CompactGitOutput` opt-out for dedicated git tools. This unconditionally passes `compactGitDiffText` into the shared runner, so `compact_git_output=false` no longer prevents semantic diff compaction (the old implementation guarded this stage, and the documented plan requires the toggle to be respected). The same regression exists in the overview/hunk routes; gate the semantic stages centrally or in each route while retaining whatever hard-truncation behavior is intended.

internal/tools/compactor_search.go
dimetron, line 110: [P1] Do not bypass `GroupSearchOutput`. This now caps and drops matches whenever the array exceeds `MaxSearchTotal`, even when `group_search_output=false`; the documented contract for that setting is the fallback path without grouping/structured-result loss. Either retain the flag's opt-out semantics for the new array cap or rename/remove the setting and update its configuration contract.

internal/tools/compactor_git.go
```diff
@@ -6,125 +6,158 @@ import (
...
 	}
 
-	origSize := len(output)
-	var techniques []string
-
-	if cfg.CompactGitOutput {
-		output = runStage(output, &techniques, "git-compact", func(s string) (string, bool) {
-			return compactGitStatusText(s, cfg)
-		})
+	capped, cut := capArray(hunks, cfg.MaxDiffLines)
```
dimetron: [P1] Cap `Hunk.Content`, not just the number of hunk records. `MaxDiffLines` is a line/size budget, but `git-hunk` can return one hunk with arbitrarily large `Content`; this path leaves that payload untouched because `len(hunks)` is still below 100. The added all-tools test only uses many small hunks, so a common single-large-hunk diff can still consume the entire result budget.

internal/tools/compactor_git.go
```diff
@@ -158,8 +200,34 @@ func (c *diffTextCompactor) flushFile() {
 	}
 }
 
+// flushHunkNote discloses the change lines dropped from the hunk just closed,
```
dimetron, line 281: [P2] Flush the per-hunk note when a new `@@` header starts. `flushHunkNote` is only called at file boundaries and at the final cut, so dropped counts from an earlier hunk are carried into later hunks and the disclosure is emitted after the last hunk instead of next to the hunk that was truncated. Call it (and reset the per-hunk counters) before resetting `hunkLines` on each new hunk header.

dimetron (commented):
Found four actionable issues in the compaction changes. Targeted compactor tests pass locally.

— Codex review

dimetron:
## govulncheck

Run after the review was resolved, against `28d7698`.

```
$ make vulncheck
10 finding(s) with no fix available — reported, not failing:
  GO-2025-3557  github.com/ollama/ollama
  GO-2025-3558  github.com/ollama/ollama
  GO-2025-3559  github.com/ollama/ollama
  GO-2025-3582  github.com/ollama/ollama
  GO-2025-3689  github.com/ollama/ollama
  GO-2025-3695  github.com/ollama/ollama
  GO-2025-3824  github.com/ollama/ollama
  GO-2025-4251  github.com/ollama/ollama
  GO-2026-5750  github.com/ollama/ollama
  GO-2026-5932  golang.org/x/crypto

```

**Gate result: pass.** No finding names a fixed version, so there is nothing to upgrade to — the gate deliberately fails only on actionable findings, since a permanently red check is one nobody reads.

All 10 are pre-existing dependency advisories in `ollama` and `x/crypto`, unrelated to this PR: it changes no `go.mod`/`go.sum` and adds no dependency. The scan is reported for the branch state rather than as a consequence of these changes.

dimetron:
Scanner versions for the run above: `govulncheck@v1.8.0`, Go `go1.27.1`, DB `https://vuln.go.dev` updated `2026-09-16 18:00:43 UTC`. The answer is only true for the database on the day it ran.

internal/cli/interactive.go
```diff
@@ -580,9 +580,12 @@ func buildDeferredCallbacks(
...
 	}
-	// Dedup runs after the compactor so both calls are compared in their final,
-	// post-compaction form.
-	afterCBs = append(afterCBs, compactorCB, tools.BuildDedupCallback(resultDeduper))
+	// Dedup runs BEFORE the compactor: its hash must cover the bytes the tool
+	// produced, not the truncated form. Compaction is lossy, so hashing after it
+	// makes two different results collide and the second is wrongly reported as
+	// "content is unchanged". See the note in cli.go where the same chain is
+	// wired for non-interactive runs.
+	afterCBs = append(afterCBs, tools.BuildDedupCallback(resultDeduper), compactorCB)
```
dimetron, line 588: ADK's `Flow.invokeAfterToolCallbacks` stops at the first callback that returns a non-nil result. The tracing callback appended just above always returns `result`, and both callbacks added here do too, so production never reaches this dedup/compactor pair (and if it did reach dedup, dedup would stop the chain before compactor). The new test manually invokes every callback and therefore misses the runtime behavior. Please compose these stages into one callback, or make intermediate callbacks return nil while preserving the final result, and test through ADK's stop/continue semantics. The non-interactive chain has the same issue below.

internal/cli/cli.go
```diff
@@ -798,13 +798,17 @@ func runNonInteractive(
...
+	// report a changed file as "content is unchanged". The deduper's contract is
+	// that the model is never served stale bytes, which only holds when the hash
+	// covers the pre-compaction content.
 	resultDeduper := tools.NewResultDeduper()
 	afterCBs = append(afterCBs,
 		lsp.BuildLSPAfterToolCallback(lspMgr),
-		tools.BuildCompactorCallback(compactorConfigFrom(cfg), tools.NewCompactMetrics()),
-		tools.BuildDedupCallback(resultDeduper))
+		tools.BuildDedupCallback(resultDeduper),
+		tools.BuildCompactorCallback(compactorConfigFrom(cfg), tools.NewCompactMetrics()))
```
dimetron, line 811: This non-interactive chain has the same ADK short-circuit bug: `invokeAfterToolCallbacks` returns immediately on a non-nil callback result, while the tracing callback and both callbacks here return `result`. Consequently the intended dedup-before-compactor order is not executed in this mode; the first earlier callback wins and the later stages are skipped. Please use a composed callback (or nil-returning intermediates) and add a test that exercises the actual ADK callback contract.

internal/tools/compactor_git.go
```diff
@@ -6,125 +6,236 @@ import (
...
 	}
 
-	origSize := len(diff)
-	var techniques []string
-
-	if cfg.CompactGitOutput {
-		diff = runStage(diff, &techniques, "git-compact", func(s string) (string, bool) {
-			return compactGitDiffText(s, cfg)
-		})
+	capped, cut := capArray(commits, cfg.MaxLogEntries)
```
dimetron, line 66: `git-overview` cannot produce an array this large: `collectGitHead` calls `git log --oneline -10`, so `recent_commits` is at most 10 items, while `DefaultCompactorConfig.MaxLogEntries` is 40. With the production defaults `cut` is therefore always false and this new pipeline is a no-op; the 200/400-commit fixtures in the tests are impossible producer shapes. Please align the producer and cap (or compact a genuinely unbounded field) and bound the test fixture by what the tool can actually emit.

dimetron (commented):
— Codex review

dimetron (commented):
Two independent reviews, both verified against the code before acting.

**Copilot found four issues; all four reproduce.**

**1. [CRITICAL] The after-tool chain stops at its first callback, so dedup, LSP and the compactor never run.** ADK's `Flow.invokeAfterToolCallbacks` (`adk/v2@v2.4.0 internal/llminternal/base_flow.go:1436`) returns at the first callback yielding a non-nil result. Every pi-go callback returns the result map, and `tracingAfter` is registered first and always returns non-nil — so the chain ends there. I proved this with a real ADK agent run: two callbacks wired, only the first fired, and the second's mutation never reached the model (25600 bytes arrived untruncated). Switching both to `return nil` (mutate in place) made both run and the change land (331 bytes). **This is pre-existing, not introduced here, and it makes this PR's savings unrunnable rather than wrong.** Notably the fix already exists in-tree — `composeAfterTool` (`piagent/callbacks.go`), with passing tests — but only `piagent/agent.go` calls it; `internal/cli/cli.go`, `internal/cli/interactive.go` and `internal/acp/server/runtime.go` still hand ADK a raw slice.

**2. [HIGH] `TestCompactGitOverview_CapsCommitsNotFileLists` feeds 200 synthetic commits** (`internal/tools/compactor_test.go:1201`), an input the tool cannot produce — `git log --oneline -10` (`git_overview.go:72`) caps it at 10. This is the exact anti-pattern this PR's own AGENTS.md section calls out, so the fixture should be bounded by the producer.

**3. [MEDIUM] Dedup is inert for two of the three git tools it lists.** `primaryOutputField` (`internal/tools/dedup.go:166`) recognises only string fields (`stdout|content|output|diff|result|data`). `git-hunk` carries `hunks` (an array) and `git-overview` has none of those keys, so `dedupTools` listing them is misleading; only `git-file-diff` actually dedups.

**4. [MEDIUM] This PR's own ordering test masks the semantics it claims to test.** `runChain` in `internal/cli/callback_order_test.go` feeds every callback's output forward regardless of nil-ness — the opposite of ADK's short-circuit — so it passes while the real chain would stop at the first entry. A guard that cannot fail is worthless, which is exactly what my AGENTS.md section warns about.

**Codex found the same short-circuit independently** in an earlier pass on this PR (its first review), which is what prompted the investigation above.

I verified finding 1 by running a real agent, not by reading signatures. Findings 2-4 are confirmed by reading the cited lines. I have not yet applied the fixes; they are out of this PR's original scope and one needs a decision (compose at all four sites vs. fix the contract).

— Codex review

internal/tools/compactor_test.go
```diff
@@ -988,122 +1037,281 @@ func TestCompactRead_SmallContent(t *testing.T) {
...
+		t.Errorf("wrote %q, want diff", cr.Writes[0].Key)
 	}
 }
 
-func TestCompactGitHunk_DiffField(t *testing.T) {
+func TestCompactGitOverview_CapsCommitsNotFileLists(t *testing.T) {
 	cfg := DefaultCompactorConfig()
-	result := map[string]any{"diff": generateLargeDiff()}
-	cr := compactGitHunk(result, cfg)
+	commits := make([]string, 200)
```
dimetron: This fixture feeds 200 commits to a tool that can only produce 10 (`git_overview.go:72`, `git log --oneline -10`), so it measures an impossible input — the same anti-pattern this PR's new AGENTS.md section documents. Bound it by the producer, or note explicitly that it exercises a shape production cannot emit.

internal/tools/dedup.go
```diff
@@ -101,9 +101,15 @@ func (d *ResultDeduper) Reset() {
 }
 
 // BuildDedupCallback returns an AfterToolCallback that elides byte-identical
-// repeat results. It must run AFTER the compactor so that both calls are
-// compared in their final, post-compaction form — otherwise two results that
-// compact to the same bytes would still be sent twice.
+// repeat results.
```
dimetron, line 109: The doc comment here is correct, but `dedupTools` still lists `git-hunk` and `git-overview`, and `primaryOutputField` only recognises string fields — so neither can ever dedup (`git-hunk` carries the array `hunks`; `git-overview` has no recognised key). Either serialise array payloads for hashing or drop them from the list, so dedup coverage is not overstated.

internal/cli/callback_order_test.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	t.Cleanup(func() { _ = sandbox.Close() })
+
+	base := buildDeferredCallbacks(config.Config{}, "anthropic", sandbox, nil, nil)
+	tool := &namedToolStub{name: "git-file-diff"}
+	args := map[string]any{"file": "big.go"}
+
+	// Long shared head (so the compacted result clears the dedup size floor),
+	// then 4000 lines that differ between the two calls and sit past the cap.
+	mkDiff := func(tag string) map[string]any {
```
dimetron, line 53: `runChain` feeds every callback's output into the next regardless of nil, which is the opposite of ADK's semantics: `Flow.invokeAfterToolCallbacks` returns at the first non-nil result. The test therefore passes even though the real chain stops at the first callback. Model the short-circuit (or compose the chain) or this guard cannot fail.

dimetron (commented):
Second part of the review (the first covered the impossible 200-commit fixture).

**1. [CRITICAL] The after-tool chain stops at its first callback, so dedup, LSP and the compactor never run.** ADK's `Flow.invokeAfterToolCallbacks` (`adk/v2@v2.4.0 internal/llminternal/base_flow.go:1436`) returns at the first callback yielding a non-nil result. Every pi-go callback returns the result map, and `tracingAfter` is registered first and always returns non-nil, so the chain ends there. I proved this by running a real ADK agent with two after-tool callbacks: only the first fired and the second's mutation never reached the model (25600 bytes arrived untruncated). Switching both to `return nil` (mutate in place) made both run and the change land (331 bytes).

This is **pre-existing, not introduced by this PR**, and it means the PR's savings cannot execute rather than being wrong. The fix already exists in-tree: `composeAfterTool` (`piagent/callbacks.go`) folds the chain into one callback, with passing tests — but only `piagent/agent.go` uses it. `internal/cli/cli.go`, `internal/cli/interactive.go` and `internal/acp/server/runtime.go` still hand ADK a raw slice, so they remain affected.

**2. [MEDIUM] Dedup is inert for two of the three git tools it lists.** `primaryOutputField` (`internal/tools/dedup.go`) recognises only string fields (`stdout|content|output|diff|result|data`). `git-hunk` carries `hunks` (an array) and `git-overview` has none of those keys, so `dedupTools` naming them is misleading; only `git-file-diff` actually dedups.

**3. [MEDIUM] This PR's own ordering test masks the semantics it claims to test.** `runChain` in `internal/cli/callback_order_test.go` feeds every callback's output forward regardless of nil-ness — the opposite of ADK's short-circuit — so it passes while the real chain would stop at the first entry. A guard that cannot fail is worthless, which is what this PR's own AGENTS.md section warns about.

Codex reached the same short-circuit conclusion independently in its earlier review on this PR, which is what prompted the investigation.

Verified by running a real agent (finding 1) and by reading the cited lines (findings 2-3). No fixes applied yet: they are outside this PR's original scope and finding 1 needs a decision about where to compose.

— Codex review

PR #371, opened 2026-09-20
feat(piagent): wire up end-of-session summaries

dimetron:
## What

The session-summary pipeline existed in full and had **no caller**.

In `~/.pi-go/memory/claude-mem.db`: **7151 sessions, 0 completed, 1 observation,
0 summaries**. Nothing ever called `SummarizeSession`, `UpsertSummary` or
`CompleteSession` outside tests. `git log --all -S <symbol>` over every commit
that ever contained each symbol shows the non-test caller count has always been
zero — the summarizer was written and never connected.

The gap was not one missing call. `Store` had no way to read back a single
session's observations either: `RecentObservations` is project-scoped.

## Changes

| Change | Where |
|---|---|
| `Store.SessionObservations(sessionID)` — one session's observations, oldest first | `internal/memory/store.go` |
| `memory.LLMSummarizer` — summary from a `model.LLM`, in process | `internal/memory/summarize.go` |
| `WithSessionSummary(m)` — opt in (memory already on) | `piagent/options.go` |
| `Agent.SummarizeSession(ctx, id)` — explicit, for your own timing | `piagent/agent.go` |
| `Agent.Close` — summarizes then completes each session it created | `piagent/agent.go`, `piagent/build.go` |

## Two decisions worth reviewing

**1. The summarizer does not spawn the `memory-compressor` subagent.**

The obvious wiring is to reuse `SubagentCompressor.SummarizeSession`. I did not,
because `specs/memory-fixes/research/findings.md` F2 measures **5.6s per spawned
observation against a 5s shutdown budget**, and records the resulting
`sql: database is closed` errors. Building on that would ship a feature with
recorded evidence it does not fit. A single in-process `GenerateContent` is one
round trip and leaves nothing orphaned when the parent exits.

**2. Ordering inside the memory closer is load-bearing — and my first draft got
it wrong.**

It must be: **drain the observation worker → summarize → close the store.**
My initial version summarized at the top of `Close()`, before `closeMemory()`
ran. Since `Shutdown` is what empties the worker's queue into the store, that
version silently dropped whatever the session's last tool calls had queued —
i.e. the end of the session, which is exactly what a summary is for. A test now
pins the order, and reverting it reproduces F2's `sql: database is closed`.

## Behaviour

- Sessions are marked **completed even with no summary model configured.** An
  `active` row is indistinguishable from a live session.
- `WithSessionSummary` is deliberately distinct from `WithSummarizer`
  (compaction). Different questions, different stores.
- With memory off, no store, no summary, no model call.
- Summarization on `Close` is best-effort, bounded by a 20s budget, and a
  failure never becomes a `Close` error.

## Verification

- `go build ./...` — clean.
- `go test -race ./internal/memory/... ./piagent/... ./internal/tools/` — pass.
- `golangci-lint run` on all four touched packages — **0 issues**.
- `go vet` on all touched packages — clean.

**Falsification** — each new guarantee was checked by breaking the code and
confirming the test fails:

| Broken on purpose | Test result |
|---|---|
| Remove the flush from `Close` | ❌ fails: "summarizer was never called", "got 0 summaries" |
| Summarize before the worker drains | ❌ fails: "saw 0 observations, want 3" |
| Close the store before summarizing | ❌ fails: "store was already closed" |
| Hold the summarizer with memory off | ❌ fails: "held despite memory being off" |
| Drop `ORDER BY`/`WHERE` in `SessionObservations` | ❌ fails: cross-session rows leak, order wrong |
| `INSERT` instead of `INSERT OR REPLACE` | ❌ fails: second call violates UNIQUE |

`internal/cli`'s **full** suite hangs in this sandbox (`httptest` cannot bind).
Confirmed pre-existing: the pristine tree exits 124 with zero output. Its
memory-specific tests pass (`ok ... 3.2s`).

## Note for the maintainer

This branch is based on `origin/main` because local `main` is 19 commits ahead
and 14 of those (`PDD plan: features/TOO/026-async-input-subagent`,
`specs/features/TOO/026-async-input-subagent/*`) are unsigned, which the
`pre-push` hook rejects. They are unrelated to this change (zero file overlap),
so I rebased onto `origin/main` rather than re-sign someone else's work. Those
14 commits still need re-signing before local `main` can be pushed.


piagent/build.go
```diff
@@ -201,9 +219,20 @@ func setupMemory(ctx context.Context, o options, cfg config.Config, orch *subage
...
+		// whatever the session's last tool calls had queued — and a summary that
+		// misses the end of a session is wrong in the way that is hardest to
+		// notice. The store then has to outlive the summary, because that is
+		// where the summary is written.
 		shutdownCtx, cancel := context.WithTimeout(context.Background(), memoryShutdownTimeout)
-		defer cancel()
 		_ = worker.Shutdown(shutdownCtx)
+		cancel()
+
+		summarize()
```
dimetron: [P1] Do not continue after a timed-out worker drain

`Worker.Shutdown` returns a timeout while its worker goroutine is still able to call `InsertObservation`, but the error is discarded before this summary hook and the subsequent `store.Close()`. That can summarize only a prefix of the session and race an in-flight insert against a closed database. This is reachable with the existing `SubagentCompressor` path (the branch documents work taking longer than the 5s drain budget). Handle the timeout before summarizing/closing, or make shutdown guarantee that the worker has stopped before the store is closed.

piagent/agent.go
```diff
@@ -324,6 +350,76 @@ func (a *Agent) Close() error {
...
+//
+// The context is bounded separately from the caller's, because Close may be
+// reached from a canceled context — that is the common case for a Ctrl-C — and a
+// summary that inherits that cancellation would never be written.
+func (a *Agent) summarizeSessions() {
+	if a.memStore == nil || len(a.sessions) == 0 {
+		return
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), sessionSummaryTimeout)
```
dimetron: [P1] Keep completion independent of the summary deadline

The same 20-second context is reused for every session in the loop and is also passed to `CompleteSession`. If one provider call consumes the budget, later sessions immediately fail, and the completion call receives an already-canceled context, leaving the session `active`; `a.sessions` is then cleared so a later `Close` cannot retry. Use a per-session summary deadline and a fresh context for completion, and retain failed session IDs for retry or explicit reporting.

piagent/agent.go
```diff
@@ -341,6 +437,7 @@ func (a *Agent) NewSession(ctx context.Context) (string, error) {
 			StartedAt: time.Now(),
 			Status:    "active",
 		})
+		a.sessions = append(a.sessions, sessionID)
```
dimetron: [P1] Synchronize session bookkeeping

`Agent` documents that it is safe for concurrent use across sessions, but this unsynchronized append races with another `NewSession` (and with `Close` reading/clearing the slice). Concurrent callers can lose session IDs or corrupt the slice, causing summaries/completions to be skipped. Protect the list with a mutex and snapshot it under that lock before closing.

internal/memory/store.go
```diff
@@ -148,6 +149,26 @@ func (s *SQLiteStore) RecentObservations(ctx context.Context, project string, li
...
+// SessionObservations returns every observation recorded for one session, in
+// the order they were recorded.
+//
+// Unlike [Store.RecentObservations] this is not project-scoped and is not
+// capped: it is what session summarization reads, and a summary that silently
+// dropped the oldest observations would describe a session that never happened.
+func (s *SQLiteStore) SessionObservations(ctx context.Context, sessionID string) ([]*Observation, error) {
+	rows, err := s.db.QueryContext(ctx,
+		`SELECT id, session_id, project, title, type, text, source_files, tool_name, prompt_number, discovery_tokens, created_at
+		 FROM observations WHERE session_id = ? ORDER BY created_at_epoch ASC`,
```
dimetron: [P2] Add a deterministic tie-breaker to observation order

`created_at_epoch` stores only whole seconds, while tool observations are commonly inserted within the same second. Rows with equal keys have undefined SQLite order, so this can violate the documented oldest-first contract and make the summary prompt reorder observations between runs. Order by `created_at_epoch ASC, id ASC` (or preserve sub-second timestamps).

dimetron (commented):
Review findings:

- The worker-drain timeout is ignored before summarization and database close, so incomplete data and closed-database races remain possible.
- A single summary context is reused across all sessions and for completion, so a slow provider can prevent later summaries and leave sessions active.
- Session bookkeeping is unsynchronized despite the Agent concurrency contract.
- Same-second observations are not deterministically ordered.

Validation: `go test ./internal/memory ./piagent ./internal/tools` passed (1675 tests); `git diff --check` passed.

— Codex review

piagent/build.go
dimetron: Fixed in 33697b1. The drain error is now checked before summarizing; on a timeout the summary is skipped with a warning rather than describing a session prefix. Worth noting this finding was load-bearing for a second reason: chasing the same shutdown path revealed that CreateSession was losing writes to SQLITE_BUSY (see the reply on the ordering finding). The first version of my concurrency test failed with "session not found" after Close, which was this bug, not a test artifact.

piagent/agent.go
dimetron: Fixed in 33697b1. Each summary now gets its own deadline and CompleteSession gets a fresh 5s context (sessionCompleteTimeout), so a summary overrun can no longer leave a session 'active'. New test TestCompletionSurvivesASummaryOverrun drives 3 sessions with a failing summarizer and asserts all 3 are completed and the summarizer was consulted 3 times.

piagent/agent.go
dimetron: Fixed in 33697b1 — and this was the most valuable of the four. -race confirmed the race on the append. Fixing it exposed a pre-existing data-loss bug: OpenDB set PRAGMAs on the pool once, so only 1 of 8 pooled connections had busy_timeout (the rest were 0), and concurrent writers lost writes with SQLITE_BUSY — 5 of 12 concurrent CreateSession calls persisted while reporting success. That is why my test reported a session 'not found' after Close: the row was never written. Pragmas now ride on the DSN, applied per connection; verified all 8 connections report busy_timeout=5000 and 12 of 12 creates persist.

internal/memory/store.go
dimetron: Fixed in 33697b1, with a correction to how I first checked it.

My initial probe passed, and that was a bad verification: I wrote a same-second test, saw it pass, and nearly concluded the finding was theoretical. Undefined order is not always the wrong order, so passing proved nothing. A second probe that fragments rowids (delete a middle run, re-insert with earlier timestamps) reproduced genuine reordering — [o-0 o-1 o-2 o-3 o-7 o-8 o-9 o-4 o-5 o-6]. The sort is now ORDER BY created_at_epoch ASC, id ASC.

Two things I want to be explicit about rather than overclaim:

1. Sub-second timestamps are not persisted (created_at_epoch is whole seconds), so within one second the order is by insertion, not by CreatedAt. The doc comment states this rather than implying timestamp-precision ordering.

2. My first replacement test asserted insertion order and still passed after I removed the tie-breaker from a scratch copy — because with an intact rowid sequence, rowid order coincides with insertion order. So that test does not pin the tie-breaker. The test kept is TestSessionObservations_SameSecondIsStableAndOrdered (stable + ids ascending + repeatable), which is what the contract actually promises. I would rather say the tie-breaker is defended by reasoning and the fragmented-rowid probe than claim a test guards it when one does not.

dimetron: Correction to my reply above, in 118c5a9.

I said the kept test did not pin the tie-breaker. That was right, and I have now fixed it rather than leaving the gap documented. Rewrote the test to put the table into the state where the two candidate orders actually diverge — insert, delete a middle run, re-insert, so the re-inserted rows carry earlier timestamps but higher ids — and to assert the contract that does hold: total order (ids ascending), no duplicates, repeatable across runs.

One limit I am recording rather than glossing over: this still cannot distinguish the explicit tie-breaker from SQLite's incidental rowid order, because an index scan over session_id visits rows in rowid order anyway. So the test pins the observable contract; the guarantee rests on reasoning. Adding uid=501(dimetron) gid=20(staff) groups=20(staff),12(everyone),61(localaccounts),79(_appserverusr),80(admin),81(_appserveradm),98(_lpadmin),399(com.apple.access_ssh),503(sandvault-dimetron),702(com.apple.sharepoint.group.2),33(_appstore),100(_lpoperator),204(_developer),250(_analyticsusers),395(com.apple.access_ftp),398(com.apple.access_screensharing),400(com.apple.access_remote_ae),703(com.apple.sharepoint.group.3),701(com.apple.sharepoint.group.1) to the ORDER BY remains correct — it removes the dependence on that incidental behaviour and on the query plan changing underneath us.

I also had to drop an assertion I had written twice: that insertion order is recoverable. It is not. created_at_epoch holds whole seconds and the sub-second component is never persisted, so once rows tie the table genuinely cannot say which came first. The first version of this test passed only because its setup made the two orders coincide.

dimetron:
## `govulncheck` — pass

```
make vulncheck   # govulncheck -format json ./... | go run ./hack/vulngate
```

| | |
|---|---|
| Scanner | `govulncheck@v1.8.0` |
| Go | go1.27.1 darwin/arm64 |
| DB | https://vuln.go.dev |
| DB updated | 2026-09-16 18:00:43 UTC |
| Branch | `feat/piagent-session-summary` @ 118c5a9 |

**Result: no findings with an available fix.**

The gate is narrow by design — it fails only on findings that name a fixed
version, because those are the ones someone can act on. This branch introduces no
new dependencies, so the result reflects the existing module graph.

10 findings have no fix available yet and are reported without failing:

| ID | Module |
|---|---|
| GO-2025-3557 | `github.com/ollama/ollama` |
| GO-2025-3558 | `github.com/ollama/ollama` |
| GO-2025-3559 | `github.com/ollama/ollama` |
| GO-2025-3582 | `github.com/ollama/ollama` |
| GO-2025-3689 | `github.com/ollama/ollama` |
| GO-2025-3695 | `github.com/ollama/ollama` |
| GO-2025-3824 | `github.com/ollama/ollama` |
| GO-2025-4251 | `github.com/ollama/ollama` |
| GO-2026-5750 | `github.com/ollama/ollama` |
| GO-2026-5932 | `golang.org/x/crypto` |

Nine are in `github.com/ollama/ollama` and one in `golang.org/x/crypto`. All
predate this branch and none are in a path this change touches; there is nothing
to upgrade to, so there is nothing to do until upstream releases a fix.

The answer is only true for the database as of the date above.

`make check-cve` was deliberately not used: it opens with `go mod tidy -v`, which
rewrites tracked files, and a check should not mutate the tree it is checking.
`git status` is clean after this run.

dimetron:
## Coverage — added code: 16 uncovered lines → 1

Measured by intersecting the diff against the coverage profile, rather than reading
a package percentage (which moves for reasons unrelated to a change).

### Codecov's verdict

| Check | Result |
|---|---|
| `codecov/patch` | ✅ **95.49% of diff hit** (target 91.37%) |
| `codecov/project` | ✅ **91.37% (+0.00%)** — no regression |

### Per-file, files this PR touches

| File | Before | After |
|---|---|---|
| `internal/memory/summarize.go` | 94% | **100%** |
| `piagent/callbacks.go` | 100% | **100%** |
| `piagent/options.go` | 100% | **100%** |
| `piagent/agent.go` | 87% | **90%** |
| `piagent/build.go` | 87% | **88%** |
| `internal/memory/store.go` | 80% | **82%** |

The gaps were all error and degraded branches. The ones worth calling out are the
branches the review produced:

- **The drain-timeout decision (P1) had no test at all** — reaching it meant
  staging a real 5s timeout. Extracted as `summarizeAfterDrain` so the choice is
  directly testable. A timed-out drain must *skip* the summary; a partial summary
  reads as complete, which is the failure mode that finding was about.

- **Store-failure paths in `SummarizeSession`** — reads failing, writes failing, a
  failing store during `Close`. A failure reported as success is worse than a
  visible one, so each asserts the error names its step.

- **A metadata-only stream frame**, which a real provider emits. Without the skip
  the summarizer would panic on the exit path.

### One line left uncovered, deliberately

`sql.Open`'s error return in `OpenDB`. It is **unreachable**: modernc's `sql.Open`
validates nothing and defers failures to first use. I confirmed this by trying
empty, whitespace, NUL-containing and malformed DSNs — none error. Left as-is
rather than adding an injected opener purely to cover a line.

### A test I had to rewrite, because it was testing nothing

My first version of the "does not repeat work" test asserted that a second `Close`
does not re-summarize. It passed — but it would have passed with the guarantee
removed, because `Close` nils its closer list, so a second `Close` never reaches
the memory closer at all. I reworked it to call `summarizeSessions` directly, then
falsified it by dropping the snapshot-clear and confirmed it fails:
`sessions are re-summarized`. The earlier version was checking the wrong mechanism.

dimetron:
## ⚠️ CI caught a regression from my own change — fixed in `dd625f6`

The first run of this commit failed the Linux `Test` job:

```
--- FAIL: TestLazyMemoryStoreDelegation (0.09s)
    memory.OpenDB: memory: migrate: record migration 1:
      constraint failed: UNIQUE constraint failed: schema_versions.version (2067)
```

**Cause, and it was mine.** When I moved the PRAGMAs onto the DSN (to fix the
`busy_timeout` bug from the review), I used `file::memory:?cache=shared` for the
`:memory:` case. That name is **process-global**, so every `OpenDB(":memory:")` in
a process resolved to the *same* database. Two concurrent opens both read schema
version 0 and both insert version 1, so one loses to the UNIQUE constraint;
migrations can also collide on the schema lock — reproduced as
`database table is locked: database is deadlocked`.

**Why it passed locally and on Windows:** it needs two in-memory opens overlapping
in time, which depends on test scheduling, not platform. I reproduced it with eight
concurrent opens.

**Fix:** each call's shared-cache name is now unique (`file:memdb-N?mode=memory&cache=shared`),
which keeps one logical database across a pool's connections — the property the
shared cache was there for — while keeping separate opens separate.

**I should have caught this myself.** My own suite runs `OpenDB(":memory:")`
concurrently. The probe I wrote for the pragma fix opened eight connections from a
*single* handle, which by construction cannot detect cross-open sharing.

Added `TestOpenDB_MemoryOpensAreIndependent`: opens concurrently, then asserts each
handle sees only its own rows. Falsified by restoring the single fixed name, which
reproduces CI's exact UNIQUE error.

Verified: build, vet, golangci-lint 0 issues, `go test -race -count=1` on
`internal/memory`, `piagent`, `internal/tools`, the memory-related `internal/cli`
tests, and CI's own race command across every package except the two known to hang
in this sandbox.

### All checks green

    ┌──────────────────────┬────────────┐
    │ Check                │ Result     │
    ├──────────────────────┼────────────┤
    │ Test                 │ ✅ pass    │
    ├──────────────────────┼────────────┤
    │ Test (Windows)       │ ✅ pass    │
    ├──────────────────────┼────────────┤
    │ Coverage             │ ✅ pass    │
    ├──────────────────────┼────────────┤
    │ Lint                 │ ✅ pass    │
    ├──────────────────────┼────────────┤
    │ Vulncheck            │ ✅ pass    │
    ├──────────────────────┼────────────┤
    │ Build ×5             │ ✅ pass    │
    ├──────────────────────┼────────────┤
    │ codecov/patch        │ ✅ 95.53%  │
    ├──────────────────────┼────────────┤
    │ codecov/project      │ ✅ 91.37%  │
    └──────────────────────┴────────────┘

